PR #10, opened 2022-05-13
Add general api

### LINKED ISSUE #5: General Script API
Features every script should have access to, i.e.:
- AddComponent
- RemoveComponent
- SpawnEntity

etc.

In the Lua case, something like this: https://github.com/PROMETHIA-27/bevy_mod_lua should work

makspll:
Closes #5 

Todo:
- [x] Move to bevy main branch for `ReflectDefault`
- [x] Complete entity/component `rlua::UserData` implementations
- [ ] Automatic support for bevy types
- [ ] Implement a base API for Rlua
- [ ] Do the same things but for Rhai


makspll:
After wrestling with this for a while, it would be very hard to pull off in bevy 0.7.
Biggest missing piece is the ability to initialise components via their `Default` implementation through `Reflect`

coderedart:
isn't it better to use mlua rather than rlua? mlua is much more actively maintained fork of rlua.

makspll:
@coderedart 
I was mostly influenced by this article: https://www.boringcactus.com/2020/09/16/survey-of-rust-embeddable-scripting-languages.html when choosing between all the lua implementations, which seemed to indicate issues using it on Windows and hence I choose rlua. However I realise that may not be the best choice, tealr seems to be a great argument for mlua since I would like to eventually be able to automatically generate documentation. Swapping to mlua should not be very difficult however since afaik the interface is really close.

lenscas:
> @coderedart I was mostly influenced by this article: https://www.boringcactus.com/2020/09/16/survey-of-rust-embeddable-scripting-languages.html when choosing between all the lua implementations, which seemed to indicate issues using it on Windows and hence I choose rlua. 

that comment about vendored mode not working with mlua on windows is not entirely correct.

On Windows, vendored mode works correctly UNLESS you use mlua to create lua libraries, something that rlua doesn't support I believe (Please correct me if I am wrong).

> However I realise that may not be the best choice, tealr seems to be a great argument for mlua since I would like to eventually be able to automatically generate documentation. Swapping to mlua should not be very difficult however since afaik the interface is really close.

tealr should work for both mlua and rlua. Please, don't let tealr alone decide what lua library you use as ideally it shouldn't have a preference for either of the 2 libraries.


> Actually, I may be able to even swap between mlua/tealr with a cargo feature.

unless I miss understand this comment, I personally don't see the point of being able to swap between mlua and tealr. Is there something that mlua does that tealr makes impossible (In which case, I missed a part of its api) or does tealr do something that is not always desirable? If either of that is true, then please let me know :)

If you mean being able to swap between mlua and rlua: Yes, their api's are similar but I'm not sure if there are many benefits to supporting both.

makspll:
@lenscas thanks for the insight!

> that comment about vendored mode not working with mlua on windows is not entirely correct.
> 
> On Windows, vendored mode works correctly UNLESS you use mlua to create lua libraries, something that rlua doesn't support I believe (Please correct me if I am wrong).

That sounds right yeah. 

> tealr should work for both mlua and rlua. Please, don't let tealr alone decide what lua library you use as ideally it shouldn't have a preference for either of the 2 libraries.

Yeah I realise this now, as of writing I had a very brief idea of tealr and was assuming it to be a wrapper around mlua. My initial thoughts in choosing between rlua/mlua were that the interfaces are similar enough that I could afford to commit to one of them and swap later if needed.

> unless I miss understand this comment, I personally don't see the point of being able to swap between mlua and tealr. Is there something that mlua does that tealr makes impossible (In which case, I missed a part of its api) or does tealr do something that is not always desirable? If either of that is true, then please let me know :)
> 
> If you mean being able to swap between mlua and rlua: Yes, their api's are similar but I'm not sure if there are many benefits to supporting both.

I misspoke here, I meant swapping `mlua/rlua`, but yes there wouldn't be a benefit to swapping here at all realy, I was just excited and thinking out loud that this would be possible :smile:.

I think ultimately mlua offers greater flexibility and I will likely end up using tealr + mlua for this reason, in fact I am performing this switch right now since I've reached a nice checkpoint on this branch. 

Ultimately LuaJit support + documentation generation are very important features.

lenscas:
just a heads up, if you use the master branch of tealr you get access to a new derive macro that might be interesting.

This new macro sets up (almost) everything needed to expose a type to lua. It however has some limitations still, which is why I haven't released it yet, but it should already work for most basic cases.

The 2 big features that are still missing is that it is unable to add an implementation for enum's with struct variants
```rs
enum Example {
    //this will not work
    notWorking {
        name:i32 
    },
    //this will work
    varant2(i32)
}
```
and it doesn't yet do anything with doc comments.

I'm more than happy to do an alpha release with the macro in this state if it makes things easier though, but ideally it gets some use before I do that :)

makspll:
Thanks for the info! that sounds great!

In my case, I am wrapping a lot of foreign types (e.g. `Vec3`, `Quat`, `Mat3` from bevy ) in `UserData` extending types. These are not however plain newtypes since they need to handle both the case in which data is owned by lua and when the data is owned by bevy (pointer to a component field) so I am not sure how much this macro would help but I'll have a look and try incorporate it if possible! 

Could you show me an example derived type?

lenscas:
I added quite a bit of documentation to this macro that you can find at https://github.com/lenscas/tealr/blob/787cc8e760945719ca7bcbd5de2b4972bfeb2bf7/src/lib.rs#L35

and there are also some tests for it at https://github.com/lenscas/tealr/blob/master/tests/mlua/to_from_derived.rs

Please let me know if something is missing from there and/or is not clear :)

PR #34, opened 2022-09-25
Add mlua/async feature support

katk0smos:
I noticed that while `mlua` supports the `async` feature (and it's bubbled up through `tealr`), this project did not support enabling it. I don't know if this was a conscious design choice, so feel free to reject this if so.


makspll:
Looks good, I don't remember excluding this feature on purpose.

Not sure what's happening with clippy here that file passed on master

PR #38, opened 2022-12-16
WIP: Update to 0.9.1

opfromthestart:
Attempting to update everything to bevy 0.9.1, and thought that it should be pushed back into the original. Not everything works yet, and I can't test teal by myself.


makefile
```diff
@@ -57,22 +57,22 @@ generate_api:
...
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_animation@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_core@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_core_pipeline@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_gltf@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_hierarchy@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_text@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_time@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_utils@0.8.1 --  -Zunstable-options --output-format json && \
-	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_reflect@0.8.1 --  -Zunstable-options --output-format json && \
+	rustup run nightly-2022-07-01 cargo rustdoc -p bevy_asset@0.9.1  --  -Zunstable-options --output-format json && \
```
makspll, line 60: Did running this work without any problems? If so that's fantastic, when I experimented with the migration initially It seemed like I'd have to bump the nighty version and do a lot of migrating

bevy_mod_scripting_core/src/lib.rs
```diff
@@ -75,11 +75,15 @@ pub trait AddScriptHost {
...
     /// Any systems which need to run the same frame a script is loaded must run after this stage.
-    fn add_script_host<T: ScriptHost, S: StageLabel>(&mut self, stage: S) -> &mut Self;
+    fn add_script_host<T: ScriptHost + Resource, S: StageLabel>(&mut self, stage: S) -> &mut Self;
 }
 
 impl AddScriptHost for App {
-    fn add_script_host<T: ScriptHost, S: StageLabel>(&mut self, stage: S) -> &mut Self {
+    fn add_script_host<T, S>(&mut self, stage: S) -> &mut Self
+    where
+        T: ScriptHost + Resource,
```
makspll, line 84: I think we can simply push this `Resource` bound into the ScriptHost definition itself to remove a lot of these explicit restrictions, It'll make everything look cleaner!

bevy_script_api/src/lua/util.rs
```diff
@@ -217,3 +229,7 @@ macro_rules! impl_tealr_generic{
 // }
 
 // pub(crate) use impl_user_data;
+
+#[derive(Resource, Default, Clone, Reflect)]
+#[reflect(Resource)]
+pub struct TypeRegistryWrapper(#[reflect(ignore)] pub TypeRegistry);
```
makspll, line 235: Do we need a wrapper? I think the type registry we should use is now [AppTypeRegistry](https://docs.rs/bevy/latest/bevy/app/struct.AppTypeRegistry.html#) or would that be inappropriate?

readme.md
```diff
@@ -326,8 +326,8 @@ To see more complex applications of this library have a look at the examples:
 - [lua - bevy API](examples/lua/bevy_api.rs)
 - [generating statically typed wrappers + ScriptRef system](examples/wrappers.rs)
 - [lua - documentation generation + lua static typing](examples/lua/documentation_gen.rs)
-- ~~[lua - bevy console integration](examples/lua/console_integration.rs)~~ not up to date with bevy 0.8
-- ~~[rhai - bevy console integration](examples/rhai/console_integration.rs)~~ not up to date with bevy 0.8
+- ~~[lua - bevy console integration](examples/lua/console_integration.rs)~~ not up to date with bevy 0.9
```
makspll, line 329: good attention to detail :) I really need to bring those examples back to life

makspll (changes requested):
Fantastic work, aligns more or less with my unfinished migration experiment, just left a few comments on things I think might benefit a discussion/improvement!

Your PR is extremely appreciated!

makspll:
Do run `cargo clippy --features=lua54,rhai,teal,lua_script_api,rhai_script_api --fix` as well please

bevy_script_api/src/lua/util.rs
opfromthestart: I was unsure how things worked, so I just did the simplest thing that worked without changing much else. I am not familiar with Reflect.

bevy_mod_scripting_core/src/lib.rs
opfromthestart: Did this.

makefile
opfromthestart: No it did not, I did not do a lot of testing of this beforehand, so much testing probably needs to occur.
With my new changes, make_json_files works, but generate_api and bevy_api_gen in general does not work.

opfromthestart:
I saw that you were attempting it through your post in the discord related to CachedScriptState, but didnt see a branch dedicated to it so just did all the changes here.

makefile
opfromthestart: I see now what you got with https://github.com/rust-lang/rust/issues/104651.  I have an idea for walking a tree but it will probably get messy and slow, is that fine?

makspll: Ah I see! So for this fix, I see two possible solutions both of which are acceptable I believe:
- One would be to simply add the import paths for each item as part of the `api_gen_config.toml` file
- Another would be to as you say walk the import tree and find the first suitable public import for each item (we only use those to generate `use` statements right now so no specific import path is necessary. Probably want all imports to have a bevy prefix as well unless there is no suitable re-export by bevy)

The first solution is obviously manual work, not insane amounts but a tad annoying, the other solution is probably better in the long run.

When it comes to speed, I think as long as the whole API generation doesn't take longer than 30-60 seconds I think I am happy with that, and as long as the `messy` methods for walking the imports are clearly separated (in their own module) it should all be fine. Let's be honest `bevy_api_gen` is a whole mess in itself rn lol

bevy_mod_scripting_core/src/lib.rs
makspll: Thanks!

makspll:
Oh I see, yeah I briefly experimented with the migration before life got in the way, I did not push these experiments to any branch, so your PR is very welcome!

bevy_script_api/src/lua/util.rs
opfromthestart: I replaced all of the wrapper types with AppTypeRegistry in my recent commit. It shows up bad since generate.rs doesn't exist.

makspll: Are you sure you pushed that commit? I cannot see relevant changes on here

opfromthestart: Ah, I may have done it with another commit, and then reverted all the changes. I'll do it soon.

makefile
opfromthestart: I have done this partially using the latter solution, but it doesn't seem to work with traits. It runs in about 2 seconds, which seems fine.

opfromthestart:
I cannot test the game of life example, which is the only one which uses a camera, so I don't know if it is working correctly, but everything appears to work now.

makefile
makspll: Looks good to me so far

makspll:
All looks fantastic, game of life works :) I'll make some changes and then merge and hopefully release soon

makspll (approved):

PR #50, opened 2023-05-08
Enable lua54 feature for docs.rs

JohnTheCoolingFan:
Fixes #49

If some other version of lua is desired for docs.rs, then tell me and I'll change it.

PR #56, opened 2023-06-01
Fix docs.rs building for bevy_mod_scripting_lua

JohnTheCoolingFan:
Same fix as in #50

Sorry, I haven't tested this on my machine yet, but it probably should work as the issue is the same and same fix here

### DIFF
--- a/languages/bevy_mod_scripting_lua/Cargo.toml
+++ b/languages/bevy_mod_scripting_lua/Cargo.toml
@@ -12,7 +12,7 @@ categories = ["game-development"]
 readme="readme.md"
 
 [package.metadata."docs.rs"]
-features = ["lua54","lua_script_api","teal"]
+features = ["lua54"]
 
 [package.metadata.release]
 pre-release-replacements = [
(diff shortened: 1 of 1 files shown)


languages/bevy_mod_scripting_lua/Cargo.toml
```diff
@@ -11,6 +11,9 @@ keywords = ["bevy", "gamedev", "scripting", "rhai"]
 categories = ["game-development"]
 readme="readme.md"
 
+[package.metadata."docs.rs"]
+features = ["lua54","lua_script_api","teal"]
```
makspll, line 15: I believe only the lua54 feature should be included as, the others are not from this crate:
```toml
# enables loading possibly unsafe lua modules by lua scripts
unsafe_lua_modules = []

# enable teal utilities
teal = []

lua51 = ["tealr/mlua_lua51"]
lua52 = ["tealr/mlua_lua52"]
lua53 = ["tealr/mlua_lua53"]
lua54 = ["tealr/mlua_lua54"]
luajit = ["tealr/mlua_luajit"]
luajit52 = ["tealr/mlua_luajit52"]
mlua_serialize = ["tealr/mlua_serialize"]
mlua_macros = ["tealr/mlua_macros"]
mlua_async = ["tealr/mlua_async"]
```

makspll, line 15:
```suggestion
features = ["lua54"]
```

makspll (changes requested):
Good catch, yeah it looks like the documentation page is not picked up by google at all for the main crate :C

makspll (approved):

PR #67, opened 2023-08-24
Proxy derive macros, rustc plugin codegen, safety improvements

makspll:
Macros of the form:

``` rust
#[reflect(Resource, LuaProxyable)]
#[proxy(languages("on_feature(lua)"), derive(Clone))]
#[functions[
    #[lua(MutatingMethod)]
    fn set_my_string(&mut self, another_string: Option<String>);

    #[lua(Method)]
    fn get_my_string(&self) -> String;


    #[lua(MetaMethod)]
    fn ToString(&self) -> String {
        format!("{:#?}", _self)
    }
]]
```


makspll:
This is almost finished, I just have codegen for binary and unary operators left over and that should be it, will also write some solid UI tests and others and voila

PR #68, opened 2023-08-24
check if entity exists before accessing

LaserWitch:
I've run into a few panics with world access if you happen to hold on to entity values between frames. I'm not certain this is the ideal way to handle them, so feel free to suggest changes or go with an alternate implementation if desired.

### DIFF
--- a/bevy_script_api/src/common/bevy/mod.rs
+++ b/bevy_script_api/src/common/bevy/mod.rs
@@ -208,11 +208,14 @@ impl ScriptWorld {
     ) -> Result<Option<ScriptRef>, ScriptError> {
         let w = self.read();
 
+        let entity_ref = w.get_entity(entity).ok_or_else(|| {
+            ScriptError::Other(format!("Entity is not valid {:#?}", entity))})?;
+
         let component_data = comp_type.data::<ReflectComponent>().ok_or_else(|| {
             ScriptError::Other(format!("Not a component {}", comp_type.short_name()))
         })?;
 
-        Ok(component_data.reflect(w.entity(entity)).map(|_component| {
+        Ok(component_data.reflect(entity_ref).map(|_component| {
             ScriptRef::new_component_ref(component_data.clone(), entity, self.clone().into())
         }))
     }
@@ -227,7 +230,10 @@ impl ScriptWorld {
             ScriptError::Other(format!("Not a component {}", comp_type.short_name()))
         })?;
 
-        Ok(component_data.reflect(w.entity(entity)).is_some())
+        let entity_ref = w.get_entity(entity).ok_or_else(|| {
+            ScriptError::Other(format!("Entity is not valid {:#?}", entity))})?;
+
+        Ok(component_data.reflect(entity_ref).is_some())
     }
 
     pub fn remove_component(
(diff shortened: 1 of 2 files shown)


bevy_script_api/src/sub_reflect.rs
```diff
@@ -348,6 +348,13 @@ impl ReflectPath {
...
                 let g = world_ptr.read();
 
+                g.get_entity(*entity).ok_or_else(|| {
+                    ReflectionError::InvalidBaseReference {
+                        base: self.base.to_string(),
+                        reason: "This entity does not exist".to_owned(),
+                    }})?;
+
+
                 let ref_ = self.walk_path(comp.reflect(g.entity(*entity)).ok_or_else(|| {
```
makspll, line 358: Just so we don't look up the entity twice
```suggestion
                let entity_ref = g.get_entity(*entity).ok_or_else(|| {
                    ReflectionError::InvalidBaseReference {
                        base: self.base.to_string(),
                        reason: "This entity does not exist".to_owned(),
                    }})?;
                let ref_ = self.walk_path(comp.reflect(entity_ref).ok_or_else(|| {
```

makspll: I am assuming you were getting panics in the `g.entity` call right?

bevy_script_api/src/common/bevy/mod.rs
```diff
@@ -169,6 +169,11 @@ impl ScriptWorld {
     ) -> Result<ScriptRef, ScriptError> {
         let mut w = self.write();
 
+        let _ = match w.get_entity(entity) {
```
makspll, line 172: can we cache the entity as in the other comment here as well?

makspll (changes requested):
Great catch! Would you be okay adding similar checks where we use `w.entity` or `w.entity_mut`. Much appreciated!

bevy_script_api/src/sub_reflect.rs
LaserWitch: Right. Good call with not doing the lookup twice, I'll incorporate that

LaserWitch:
> Great catch! Would you be okay adding similar checks where we use `w.entity` or `w.entity_mut`. Much appreciated!

Yeah, I'll do a search and see how many other places I can add it in.

I'm honestly tempted to try to implement an option or result style construct in lua, perhaps leverage multiple value return, but I think I can experiment with that a bit further down stream for the time being.

makspll:
> Yeah, I'll do a search and see how many other places I can add it in.
Thanks!

> I'm honestly tempted to try to implement an option or result style construct in lua, perhaps leverage multiple value return, but I think I can experiment with that a bit further down stream for the time being.

I've considered something like this in the past, although I've stuck with the mlua convention of converting `None` -> `nil` and `Error` -> Runtime Error, while creating proxy types for things like `Vec` when maintaining `pass-by-reference` behavior's.

LaserWitch:
I've been lost in some other weeds the past week but I'll see if I can get this done this weekend.

LaserWitch:
I have some more bases covered now. There's some other things that look like potential cases that could be checked, but I wouldn't really want to mess with them without first having a test case that is known to fail, and building those tests cases is not something I'm up for doing at the moment

makspll (dismissed):
Looks good, if we need any more checks anywhere else we can add later. Thanks!

LaserWitch:
Looks like I've finally made CI happy. Not sure why it says I dismissed the review above, that must have been something automatic.

makspll (approved):

makspll:
Yeah it's automatic, might disable that as the message looks quite confusing.

Issue #65, opened 2023-08-22
Can persistant, shared globals state be enabled?


LaserWitch:
I'm not sure if this is a FR or something else, but here goes. 

In my testing I've found global state isn't behaving how I expect Lua state in a game to act. It appears that each script runs inside it's own, enviroment/context. A global defined by one script is nil when another script tries to access it. I can see some cases for this, but it also seems to inhibit many of the approaches to scripting that I want to enable unless I go through a lot of effort to essentially reimplement parts of Lua as resources.

Perhaps relatedly, when a script is reloaded the original global state seems to be discarded and a fresh one replaces it. for instance this hook:
```lua
function on_level()
    if type(_G.i) == 'nil' then _G.i = 0 end
    if(i%200 == 0) then
        print("\n******** " .. i )
    end
    i=i+1
end
```
In theory the count should persist between reloads, if there was a single context and the hook function was simply replaced by executing the changed file. It appears that it's a clean new context though. Any stateful gameplay scripting live development is going to be really awkward this way. If I make use of `require` then it may be viable to code around this to a limited degree inside lua itself, but I'm not sure that can cover everything.

Do you think it'd be feasible for a game to get around either of those with custom ScriptHost implementations? If not, is it conceivable to make those alternate behaviors available somehow, likely optionally? I've only got a loose grasp of the internals of the crate so far, so I'm not able to make a confident guess myself.

makspll:
Hi again! all great questions!

In short, yes I believe all of this could be handled by a custom `ScriptHost`. The interface is very flexible and in fact `register_with_app` of each ScriptHost is the sole point of contact with the app for the entire crate.

Discounting this itty bitty fragment of  shared code:
bevy_mod_scripting_core/src/lib.rs
```rust
    45 | fn build(&self, app: &mut bevy::prelude::App) {
    46 |  app.add_event::<ScriptErrorEvent>();
    47 | }
```

The removal of old contexts happens in the  `reload_script` method of a script which is used by the `script_hot_reload_handler` system

bevy_mod_scripting_core/src/hosts.rs
``` rust
     366 |  // remove old context
     367 |  contexts.remove_context(script.id());
```

This system is registerd with the app in each `ScriptHost` like this

languages/bevy_mod_scripting_lua/src/lib.rs, from line 81
```rust
    fn register_with_app_in_set(app: &mut App, schedule: impl ScheduleLabel, set: impl SystemSet) {
[...]

PR #72, opened 2023-09-02
changing handle_events in the ScriptHost trait to be a mut method

LaserWitch:
Allows for flexibility in custom hosts, see #65.


makspll (approved):
Looks straight forward, thanks!

Issue #74, opened 2023-09-30
Send Priority event from ScriptWorld


zwazel:
Hi, love this crate!

I just came across a small issue.
I need to send out PriorityEvents from a rust function that is registered as a script function.
So it has only access to the ScriptWorld.

With the Normal Bevy Events I can just use world.send_event().
But I can't seem to figure out how to send out priority events with only having the ScriptWorld.

Did I miss anything?

makspll:
Hi @zwazel thanks! 
What do you mean by a rust function registered as a script function, am I correct in thinking you're trying to send out events back from the script to the bevy world? 
In this case you can do something similar to world.send_event which is defined using send_event-batch like so: 
```  rust
    /// Sends a batch of [`Event`]s from an iterator.
    #[inline]
    pub fn send_event_batch<E: Event>(&mut self, events: impl IntoIterator<Item = E>) {
        match self.get_resource_mut::<Events<E>>() {
            Some(mut events_resource) => events_resource.extend(events),
            None => bevy_utils::tracing::error!(
                    "Unable to send event `{}`\n\tEvent must be added to the app with `add_event()`\n\thttps://docs.rs/bevy/*/bevy/app/struct.App.html#method.add_event ",
                    std::any::type_name::<E>()
                ),
        }
    }
```

in this case instead of `Events` you'll simply use `PriorityEvents` like so:
```rust
    /// Sends a batch of [`Event`]s from an iterator.
    #[inline]
    pub fn send_event_batch<E: Event>(&mut self, events: impl IntoIterator<Item = E>, prio: u32) {
        match self.get_resource_mut::<PriorityEvents<E>>() {
            Some(mut events_resource) =>events_resource.events.extend(events.map(|v| EventInstance::new(v, prio))),
            None => bevy_utils::tracing::error!(
                    "Unable to send event `{}`\n\tEvent must be added to the app with `add_event()`\n\thttps://docs.rs/bevy/*/bevy/app/struct.App.html#method.add_event ",
                    std::any::type_name::<E>()
                ),
        }
    }
```

makspll:
I should probably extend the `world` API with a custom trait containing priority event dispatchers tbh

PR #76, opened 2023-10-06
Fix bug in Rhai Event Handler

### LINKED ISSUE #75: Rhai Events are overriding each other
Hi!
Could be that I completely misunderstand events in this crate.
I've implemented a function that allows me to spawn a mod.
So via script, i can just spawn other mods.
and those spawned mods then get called the "on_init" function in their script!

problem is, as soon as there is more than just one at the same time, only one of them gets the init method called.

same thing for the Update call.
I can't use Recipients::All because i need to do some additional checks before calling on_update on that entity, so i use Recipient::Entity, and loop through all of them. and here the same problem!

~~Am I misunderstanding something or is this a "design flaw"?~~
Noticed that this is a bug in the rhai implementation. Lua did not behave Like this.

zwazel:
This should fix #75 

Also added a new example for Rhai.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ homepage = "https://github.com/makspll/bevy_mod_scripting"
 keywords = ["bevy", "gamedev", "scripting", "lua"]
 categories = ["game-development"]
 readme = "readme.md"
-include= ["readme.md","/src","/examples","/assets","LICENSE"]
+include = ["readme.md", "/src", "/examples", "/assets", "LICENSE"]
 
 [[bin]]
 name = "bevy_mod_scripting_doc_gen"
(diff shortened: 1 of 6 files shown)


zwazel:
I just noticed that i also updated the Rhai version. please tell me if I should revert this change or not.

zwazel:
I also just changed it so that i no longer get errors printed out when a function can't be found in a script.

as far as i can tell, this is a similar behaviour as in the Lua solution: https://github.com/makspll/bevy_mod_scripting/blob/75902486147e48b509e07752f713bfff459dba46/languages/bevy_mod_scripting_lua/src/lib.rs#L169-L172

makspll:
Hi @zwazel PR looks good, great catch! Love the examples. Please see to the clippy errors and LGTM

zwazel:
Glad I can help with this great crate!
I currently can't go on my PC, so i'll fix the clippy errros tomorrow!

I might also take a Look at how to tackle #74

makspll:
Appreciate it a lot! no problem!

makspll (dismissed):

makspll (approved):

zwazel:
Hi, as far as I can tell the only clippy warning that I got was from lua derive. 
(I actually didnt get any locally, so kinda hard to reproduce, but that should have fixed it)

makspll:
Yeah sorry about that, the CI runs on latest rust version meaning any new clippy warnings randomly popup on old code, I need to find a way to pick the clippy version somehow

PR #77, opened 2023-10-08
Re-generate code to fix build on ARM Macs

### LINKED ISSUE #70: Incompatible types when compiling on Arm64
I am facing a similar set of errors to #61 with `bevy 0.11.2` and bevy_mod_scripting commit 8bf3796 with the `lua_script_api` flag enabled

```shell
cargo run
    Updating git repository `https://github.com/makspll/bevy_mod_scripting`
   Compiling bevy_script_api v0.3.0 (https://github.com/makspll/bevy_mod_scripting?rev=8bf3796#8bf37962)
error[E0599]: no variant or associated item named `Pairs` found for enum `LuaMetaMethod` in the current scope
   --> /Users/videah/.cargo/git/checkouts/bevy_mod_scripting-66f82db9d3a19d6f/8bf3796/bevy_script_api/src/lua/std.rs:317:45
    |
317 |         methods.add_meta_method(MetaMethod::Pairs, |ctx, s, _: ()| {
    |                                             ^^^^^ variant or associated item not found in `LuaMetaMethod`

error[E0308]: `?` operator has incompatible types
    --> /Users/videah/.cargo/git/checkouts/bevy_mod_scripting-66f82db9d3a19d6f/8bf3796/bevy_script_api/src/generated.rs:4501:17
     |
4501 |     #[languages(on_feature(lua))]
     |                 ^^^^^^^^^^^^^^^ expected `BVec4`, found `BVec4A`
     |
     = note: `?` operator cannot convert from `BVec4A` to `BVec4`
     = note: this error originates in the macro `impl_lua_newtype` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
    --> /Users/videah/.cargo/git/checkouts/bevy_mod_scripting-66f82db9d3a19d6f/8bf3796/bevy_script_api/src/generated.rs:4501:17
     |
4501 |     #[languages(on_feature(lua))]
     |                 ^^^^^^^^^^^^^^^
     |                 |
     |                 expected `BVec4A`, found `BVec4`
     |                 arguments to this function are incorrect
     |
note: associated function defined here
    --> /Users/videah/.cargo/git/checkouts/bevy_mod_scripting-66f82db9d3a19d6f/8bf3796/bevy_script_api/src/wrappers.rs:38:16
     |
38   |         pub fn new(b: $type_) -> Self {
     |                ^^^ ---------
     |
    ::: /Users/videah/.cargo/git/checkouts/bevy_mod_scripting-66f82db9d3a19d6f/8bf3796/bevy_script_api/src/generated.rs:5050:17
     |
5050 |     #[languages(on_feature(lua))]
     |                 --------------- in this macro invocation
     = note: this error originates in the macro `impl_lua_newtype` (in Nightly builds, run with -Z macro-backtrace for more info)

Some errors have detailed explanations: E0308, E0599.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `bevy_script_api` (lib) due to 3 previous errors
```

My Cargo.toml looks like this:

```toml
[package]
name = "myrrh"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

# Enable a small amount of optimization in debug mode
[profile.dev]
opt-level = 1

# Enable high optimizations for dependencies (incl. Bevy), but not for our code:
[profile.dev.package."*"]
opt-level = 3

[dependencies]
bevy = { version = "0.11.2", features = ["dynamic_linking"] }
bevy_mod_scripting = { git = "https://github.com/makspll/bevy_mod_scripting", rev = "8bf3796", version = "0.3.0", features = ["luajit", "teal", "lua_script_api"] }
```

CyanBlob:
Updating the glam and bevy versions and then re-generating `generate.rs` fixes #70. This issue seems to ultimately stem from https://github.com/bitshifter/glam-rs/issues/296. 

This should be tested on x86 platforms before merging to ensure compatibility wasn't broken.

A better solution will be to eventually generate different code for each platform via cross-compilation.


makspll:
Hi, yes once aarch64 targets get SIMD support in glam this will disappear, however that seems to require nightly rust to run together with the `core-simd` feature!

Changing these signatures does seem to fix this on arm64 however it breaks again on x86, I guess this is the other side of the problem. 

A full fix will need to somehow work on both, probably need a bit of logic in the macro code along the lines of:
 - According to the expected signature am I generating an `A` type vector (Vec4A, BVec4A etc) on a platform with no simd support? 
 - if so before returning, convert using the `From` trait to an `A`  type vector
 
However I am not sure this generalizes to all the cases where these vectors are used. Glam is quite confusing, It contains both replacement Vec4 versions for different platforms depending on simd support (the issue we're facing) but also replacement `BVec4A` and such types which do not use simd instructions on non-simd platforms. 

To be honest this seems like a portability issue in glam itself to me. If you rely on an `A` vector being returned on x86 in your code, this will break on non simd platforms. I think maybe the best fix for now will indeed be generating alternative .rs files depending on simd support, but then removing that once the issue in glam is fixed.

makspll:
Already fixed on main

PR #80, opened 2023-10-12
Fix SIMD codegen issue

makspll:


makspll:
Already fixed in main by upstream commits

PR #81, opened 2023-10-13
Fix removal of script contexts

### LINKED ISSUE #79: How do I correctly despawn Entities with scripts?
I despawn an entity (recursively) that has a script attached to it. but the script still "exists" and receives the events.

I'm currently only working with the Rhai implementation, so idk if this is a general problem or not. but its definitely not what i expect it to work like XD

zwazel:
This should fix #79.

Okay i found following issue:
the system `script_remove_synchronizer` handles removing the contexts of the script.
https://github.com/makspll/bevy_mod_scripting/blob/32eba12bc5e2d87ab05cfbf4a2dd14063d3842a2/bevy_mod_scripting_core/src/systems.rs#L94-L103

The problem here is, that `.remove_context` expects the script ID, as `context_entities` uses it as the Key.
https://github.com/makspll/bevy_mod_scripting/blob/32eba12bc5e2d87ab05cfbf4a2dd14063d3842a2/bevy_mod_scripting_core/src/hosts.rs#L267-L271

Which results in a very unreliable removing of the scripts. especially once the entities have multiple scripts and not just one.

I have updated the `script_remove_synchronizer` system, to loop through all `context_entities` and find those with a matching entity id.
The problem I see with this solution is that we have to loop through everything, as we can't break after one found match. because each entity could have multiple scripts, so we need to loop through all of them to find all.
https://github.com/makspll/bevy_mod_scripting/blob/745c20bb9e22d18761da7c50e96ee7f4766e790f/bevy_mod_scripting_core/src/systems.rs#L94-L112

### DIFF
--- a/bevy_mod_scripting_core/src/systems.rs
+++ b/bevy_mod_scripting_core/src/systems.rs
@@ -98,15 +98,13 @@ pub fn script_remove_synchronizer<H: ScriptHost>(
     for v in query.iter() {
         // we know that this entity used to have a script component
         // ergo a script context must exist in ctxts, remove all scripts on the entity
-        let mut script_ids = Vec::new();
-        for (script_id, (entity, ..)) in contexts.context_entities.iter() {
-            if entity.index() == v.index() {
-                script_ids.push(*script_id);
-            }
-        }
-
-        for script_id in script_ids {
-            contexts.remove_context(script_id);
+        let script_ids = contexts.context_entities
+                                    .iter()
+                                    .filter_map(|(script_id, (entity, ..))|  (entity == v)
+                                    .then_some(script_id))
+                                    .collect();
+        for script_id in  script_ids{
+            contexts.remove_context();
         }
     }
 }
--- a/bevy_mod_scripting_core/src/systems.rs
+++ b/bevy_mod_scripting_core/src/systems.rs
@@ -98,7 +98,6 @@ pub fn script_remove_synchronizer<H: ScriptHost>(
     for v in query.iter() {
         // we know that this entity used to have a script component
         // ergo a script context must exist in ctxts, remove all scripts on the entity
-        println!("Removing script context for entity {:?}", v);
         let mut script_ids = Vec::new();
         for (script_id, (entity, ..)) in contexts.context_entities.iter() {
             if entity.index() == v.index() {
@@ -107,19 +106,7 @@ pub fn script_remove_synchronizer<H: ScriptHost>(
         }
 
         for script_id in script_ids {
-            println!("removing script id: {}", script_id);
             contexts.remove_context(script_id);
-        }
-
-        println!("remaining contexts: {:?}", contexts.context_entities.len());
-        for (script_id, (entity, context, some_string)) in contexts.context_entities.iter() {
-            println!(
-                "script id: {}, entity: {:?}, context: {:?}, name: {:?}",
-                script_id,
-                entity,
-                context.is_some(),
-                some_string
-            );
         }
     }
 }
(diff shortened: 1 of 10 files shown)


zwazel:
I just added some changes to the context_entities.
I would love to hear some feedback from you, I personally think it makes more sense to use the entity as a key, instead of the script ID.

zwazel:
Not sure how to fix the test, as it comes from a crate `rust_out`?

```
running 7 tests
test src/lib.rs - (line 252) ... ignored
test src/lib.rs - (line 65) ... ignored
test src/lib.rs - (line 268) - compile ... FAILED
test src/lib.rs - (line 208) ... ok
test src/lib.rs - (line 117) ... ok
test src/lib.rs - (line 139) ... ok
test src/lib.rs - (line 179) ... ok

failures:

---- src/lib.rs - (line 268) stdout ----
error[E0601]: `main` function not found in crate `rust_out`
  --> src/lib.rs:287:2
   |
21 | }
   |  ^ consider adding a `main` function to `src/lib.rs`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0601`.
Couldn't compile the test.

failures:
    src/lib.rs - (line 268)

test result: FAILED. 4 passed; 1 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.71s
```

bevy_mod_scripting_core/src/systems.rs
```diff
@@ -173,21 +173,26 @@ pub fn script_event_handler<H: ScriptHost, const MAX: u32, const MIN: u32>(world
...
-            Some((
-                ScriptData {
-                    sid: *sid,
-                    entity: *entity,
-                    name,
-                },
-                ctx,
-            ))
-        });
+        .filter_map(|(entity, vec)| {
```
makspll, line 176: what's the purpose of the sort and .flatten here?

makspll (changes requested):
While I see the benefits of keeping entities as the keys in change detection, it adds load to the hot path, i.e. iterating through all the scripts on top of decreasing cache locality with a nested Vec. 

I'd much rather have to iterate through all the contexts to find the right scripts to remove than have to do some additional work when handling events :)

bevy_mod_scripting_core/src/systems.rs
zwazel: you're correct about the sort, that has no purpose (removed in the latest commit).
Regarding the flatten, handle_events expects a `ctxs: impl Iterator<Item = (ScriptData<'a>, &'a mut Self::ScriptContext)>`
Without flattening it, `ctx_iter` has iterators in iterators, which it does not expect.
image

But I'll be fully honest with you, that part was mostly written with the help of Github Copilot, so if you got a better way / preferred way of doing so, i'm open to feedback!

zwazel:
I see, haven't considered that!
In that case I'll quickly revert back to how it was before!

zwazel:
Merge Request is back at only fixing the bug in removing script context

bevy_mod_scripting_core/src/systems.rs
```diff
@@ -95,11 +95,20 @@ pub fn script_remove_synchronizer<H: ScriptHost>(
...
+        let mut script_ids = Vec::new();
+        for (script_id, (entity, ..)) in contexts.context_entities.iter() {
+            if entity.index() == v.index() {
+                script_ids.push(*script_id);
+            }
+        }
+
+        for script_id in script_ids {
+            contexts.remove_context(script_id);
+        }
```
makspll, line 110:
```suggestion
        let script_ids = contexts.context_entities.iter().filter_map(|(script_id, (entity, ..))|  (entity == v).then_some(script_id));
        for script_id in  script_ids{
            contexts.remove_context();
        }
```

How about something like this?

makspll:
No worries, much appreciated! Just one code style comment and I am happy!

bevy_mod_scripting_core/src/systems.rs
zwazel: yep, that's better :)
Updated in the recent commit.

PR #87, opened 2023-11-10
Small cleaning up + Fix for crash when looping through empty Vec

### LINKED ISSUE #86: Crash when trying to loop through empty Vec
I got following struct:
```rust
#[derive(Resource, Reflect, Default, Clone)]
#[reflect(Resource)]
pub struct LobbyListResource {
    pub lobbies: Vec<u64>,
}
```
I register all the needed things to correctly access the vec and loop through it via scripts:
```rust
struct HandleLobbyApiProvider;

impl APIProvider for HandleLobbyApiProvider {
    type APITarget = Engine;
    type ScriptContext = RhaiContext;
    type DocTarget = RhaiDocFragment;

    fn register_with_app(&self, app: &mut App) {
        app.register_foreign_rhai_type::<Vec<u64>>();
    }

    fn attach_api(
        &mut self,
        api: &mut Self::APITarget,
    ) -> Result<(), bevy_mod_scripting::prelude::ScriptError> {
        api.register_vec_functions::<u64>();
        Ok(())
    }
}
```
With this I can then loop through the lobbies in Rhai like this:
```rhai
let lobby_list_type = world.get_type_by_name(blablabla::LobbyListResource");
if world.has_resource(lobby_list_type) {
    let lobbies = world.get_resource(lobby_list_type).lobbies;
    for lobby in lobbies {
        print(`lobby: ${lobby}`);
    }
}
```
And this works... as long as the vec is not empty!
When it's empty i get following error and it crashes:
```txt
thread 'main' panicked at 'attempt to subtract with overflow', C:\...\.cargo\git\checkouts\bevy_mod_scripting-ff78cea4271e6409\8879577\bevy_script_api\src\common\std.rs:136:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in exclusive system `bevy_mod_scripting_core::systems::script_event_handler<bevy_mod_scripting_rhai::RhaiScriptHost<pgc::modding::ModdedScriptRuntimeArguments>, 0, 10>`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```

As I'm only using Rhai, I don't know if this error exists in Lua too.

The current workaround is to just check if the vec is empty before trying to loop through it:
```rhai
let lobby_list_type = world.get_type_by_name(blablabla::LobbyListResource");
if world.has_resource(lobby_list_type) {
    let lobbies = world.get_resource(lobby_list_type).lobbies;
    if !lobbies.is_empty() {
        for lobby in lobbies {
            print(`lobby: ${lobby}`);
        }
    }
}
```

zwazel:
This pull request cleans the Rhai side of the project up a bit.
- Removed a second `get_children` that didn't even work
- fixed the `get_parent`, before it wasn't found by Scripts, presumably because not the correct World was asked for (`&ScriptWorld` instead of `ScriptWorld`). Also made it return a Entity if a parent was found, and a i32 of value -1 if not, instead of Option<Entity>, as that isnt really supported by Rhai.
- added an additional check for accessing an entity in the `reload_script` as the game often crashed there. This adds on #68 
- Also made a _partial_ fix for #86. As the app no longer crashes, but we still get an error in the console (better documented in the issue itself).
  - is now fully fixed
  - closes #86


bevy_mod_scripting_core/src/hosts.rs
```diff
@@ -360,22 +360,25 @@ impl<T: Asset> Script<T> {
...
+                entity,
+                script_assets,
+                providers,
+                contexts,
+                event_writer,
+            );
+        } else {
+            // remove old context
+            contexts.remove_context(script.id());
+        }
```
makspll, line 381: Good catch!

bevy_script_api/src/common/std.rs
```diff
@@ -131,10 +131,11 @@ impl<T: FromReflect + TypePath> IntoIterator for ScriptVec<T> {
             // TODO?: end used to be an Option, and this check moved into the next method but
             // I am not sure if this will ever realistically fail, so if you do get this exception happening
             // hit me with an issue
+            // if len > 0, subtract 1, otherwise set to 0
             end: self
                 .len()
-                .map(|v| v - 1)
```
makspll, line 136: Hmm, I think this will still have the problem that when there are zero elements, the iterator will try to index into the vec when `start == end`.

Let's maybe rename end to len, set it to len here directly to avoid these shenanigans and then change the condition in the iterator to `start < len`

bevy_script_api/src/rhai/bevy/mod.rs
```diff
@@ -138,15 +131,19 @@ impl CustomType for ScriptWorld {
                     })
                 },
             )
-            .with_fn("get_children", |self_: &ScriptWorld, parent: Entity| {
+            .with_fn("get_parent", |self_: ScriptWorld, entity: Entity| {
+                if let Some(parent) = self_.get_parent(entity) {
```
makspll, line 135: How does this work on the rhai side? Do you have to check the type of the result every time?

We also have support for options similarly to Vec with the custom wrapper, the only thing is it needs to be registered for the specific type with  `app.register_foreign_rhai_type::<Option<T>>()` I believe.

I agree though it is a bit annoying having to register all these, so assuming that returning another type is ergonomic this may be the way to go at least with the std lib for bevy. Can we use `()` instead though ?

makspll (changes requested):
Great catches, just a few comments

zwazel:
I suggest we move the existing Wrapper example into lua, as it only really shows how to work with Lua. 
I'm working on the example for Rhai right now, so we can test the fixes

zwazel:
btw while working with the Wrappers in Rhai, i noticed that i always have to manually implement RhaiCopy, even though it's completely empty.
```rust
#[derive(Resource, Reflect, Default, Clone, Debug)]
#[reflect(Resource, RhaiProxyable)]
pub struct MyThing {
    usize: usize,
    string: String,
    array: Vec<usize>,
}

impl RhaiCopy for MyThing {}
```
Is this wanted or an oversight?

zwazel:
As i'm working on the wrapper example for rhai, i noticed following thing:
When I Reflect on RhaiProxyable, I no longer can access the values of my Struct. seen in: https://github.com/makspll/bevy_mod_scripting/pull/87/commits/c282ada927a822eb88750f5e669df53421ebe543
When I don't use RhaiProxyable, everything works fine. as seen in: https://github.com/makspll/bevy_mod_scripting/pull/87/commits/59ac5629c7559e82619949ee44ef1cace006541f

Am I misunderstanding the need for RhaiProxyable?

bevy_script_api/src/common/std.rs
zwazel: I changed the iterator to
```rust
impl<T: FromReflect> Iterator for ScriptVecIterator<T> {
    type Item = ScriptRef;

    fn next(&mut self) -> Option<Self::Item> {
        let nxt = (self.current < self.len).then(|| self.base.index(self.current));
        self.current += 1;
        nxt
    }
}
```
As suggested by you, we now check `start < len`. also `end` has been renamed to `len`.
This no longer throws an error if the vec is empty, but it also doesn't fully loop through the vec:
```text
MyThing.array: [1, 2, 3]
MyThing.array[0]: 1
Looping through MyThing.array:
MyThing.array element: 1
MyThing.array element: 2
```
As you can see, the last element is not looped through.

zwazel: I changed it back to `<=` currently, so that it still loops through all elements

bevy_script_api/src/rhai/bevy/mod.rs
zwazel: Alright, I tried it out again, I'm not sure what I did different now, but I now return a UNIT (`()`) and it works in Rhai.
I've expanded the multiple_events example to check this.
in Rhai it would look like this:
```rhai
let parent = world.get_parent(entity);
switch type_of(parent) {
    "bevy_ecs::entity::Entity" => {
        print(`Parent is an Entity`);
    },
    "()" => {
        print(`Parent doesn't exist`);
    },
}
```

zwazel: > How does this work on the rhai side? Do you have to check the type of the result every time?

Not necessarily. 
If you're not 100% if the parent exists, then this is recommended yes.
But if you're sure you can just use the value returned from `get_parent`.

examples/rhai/wrappers_rhai.rs
makspll, line 1: So the wrappers example is meant to showcase a multi-lang macro, but currently it does not support rhai hence why it was in the shared folder

What you're doing in this example is identical to the bevy_api example, i.e. using plain reflection to access bevy objects with the Bevy API.

Wrappers are a slightly different flavor of access. 

To give an example, if we have a type:
```rust
#[derive(Component, Default, Reflect)]
#[reflect(Component)]
pub struct MyType {
    pub my_string: String,
    /// let's assume this is compilable, (imagine a box surrounding this, omitting for simplicity)
    pub my_sub: MyType
}
```
and then access it in lua with:
```lua
local my_thing_type = world:get_type_by_name("MyThing");
local my_thing = world:get_resource(my_thing_type);
local my_sub = my_thing.my_sub;
local my_string = my_sub.my_string;
```
Here `my_sub` will be a `ReflectedValue` and my_string will be a concrete String thanks to the implementation of LuaProxyable on the String type which essentially uses the String type as its `proxy`.  

For some types you might need to use a different proxy type, for example when you don't own the type you're working with, in which case you need to create a newtype which wraps it and implements whatever you need and implements UserData that way.

This is how the entirety of the Bevy API is generated, we use a macro to generate wrapper newtypes which implement UserData + LuaProxyable and allow us to register this proxy with the original type.

Another reason you might need a wrapper, is if you want to have a custom UserData implementation. ReflectedValue is very basic and does not support methods by default. So this is what the `wrappers.lua` type showcases, the macro which lets you avoid a lot of boilerplate in adding custom methods to a type. 

yet another is if you want to have an 'Owned' variant of your type creatable by the script, creating such static functions is supported by the wrapper macro and lets you mix script owned values (for example Vec3::new(...) on script side) vs references to values in the world (for example a Vec3 in a component) seamlessly.

If all you need is access to basic primitives and bevy types,  then ReflectedValue should get you far, otherwise you need wrappers.

TL;DR I would keep the wrappers examples as they were since this file essentially duplicates what's already inside bevy_api.rs and in the future we'll add the rhai version to the same file (I am currently re-working the macro completely and hopefully will be able to make one for Rhai at some point)

assets/scripts/multiple_events_rhai.rhai
```diff
@@ -1,5 +1,15 @@
 fn on_init(name) {
     print(`Hello World! From "${name}" in Init`);
+
+    let parent = world.get_parent(entity);
+    switch type_of(parent) {
```
makspll, line 5: is it possible to do something like:
```rhai
if parent {
  ... access parent
} else {
  ... throw error
}
```
Instead? As in will rhai automatically convert those to truthy values here ?

makspll (changes requested):
Some more discussion, and minor changes

bevy_script_api/src/common/std.rs
zwazel: oh. yep good point, that somehow flew over my head

assets/scripts/multiple_events_rhai.rhai
zwazel: almost, you can check it like this
```rhai
let parent = world.get_parent(entity);
if parent == () {
    print(`Parent doesn't exist`);
} else {
    print(`Parent exists`);
}
```

I have updated the example accordingly

examples/rhai/wrappers_rhai.rs
zwazel: Alright, i moved the wrapper example back :)

bevy_script_api/src/common/std.rs
zwazel: this seemed to fix it btw!

zwazel:
So this pull request should be good now, applied all your suggestions. And am looking forward for the wrapper rewrite ;)

closes #86

makspll (approved):
LGTM, thanks again, great work!

Issue #89, opened 2023-12-15
Can't compile with mlua anymore


LaserWitch:
I started towards trying to do my own update to bevy 0.12(haven't not noticed the issue here recently that mentioned it being possibly coming soon), and I've found that I can't build from the repo at all now when I use the lua-api and lua54 features. I get a bunch of errors from tealr macros. tealr_derive::TypeName, ToTypename, IntoLua... I'm thinking there's been some poorly-versioned breaking changes up stream but haven't yet found a correct combination of version specifications to get back to buildable.

makspll:
Hi, I think tealr got an update, I remember there being some issues with cargo selecting unexpected versions for it, have a look at: https://github.com/lenscas/tealr_doc_gen/issues/18, 
Solution is to either:
- figure out how to make cargo respect the alpha version
- update to the newest tealr version

makspll:
W.r.t bevy 0.12 I am currently heavily invested on this branch: https://github.com/makspll/bevy_mod_scripting/tree/feature/derive_macros

basically a complete re-write of the codegen + possibly allowing consumers to generate an API similarly to how we do it for Bevy, and was hoping to finish that before releasing, but given it's taking me way longer than expected, I might just release a pure bevy patch (for both 0.11 and 0.12), and PR's are very much welcome!

LaserWitch:
Right now my own project is awkwardly pinned to a bevy 0.12-dev version to clean up some issues, so I was trying to do a 0.12 patch to be able to be solidly on released bevy version. I'll see if I can do an update to just current tealr versions to get that cleaned up and then move on to the 0.12.

lenscas:
Sorry about this. I didn't know (nor expected) that me releasing a non alpha version would break this. In fact, as far as I understand how cargo should resolve versions it should go for the exact version if you add stuff like --alpha after the version number. I guess I need to re read how this part work to prevent it from happening in the future.

Again, sorry that this happened. It was not at all my intention to break anything, and especially not without notice.

PR #90, opened 2023-12-15
update tealr dependancy

LaserWitch:
This is an attempt to get around compilation issues with lua versions. 
Unfortunately it's not working yet. It seems there's some trait changes that impact codegen that I'm not sure how to fix yet. I figured a draft PR with my partial work might still be of value to someone, at least.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,14 +59,14 @@ rhai = ["bevy_mod_scripting_rhai"]
 rhai_script_api=["bevy_script_api/rhai"]
 
 [dependencies]
-bevy = { version = "0.11", default-features = false}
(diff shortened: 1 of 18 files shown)


LaserWitch:
I have belatedly realized, after properly digesting the cause identified to me in #89, that the root issue can be handled much more expediently with a `[patch]` section in the main cargo file to point at the github revision of 0.9.0-alpha4.
It'd be good for the crate to point to a non-alpha version of tealr eventually, but perhaps that can wait for the new codegen before being tackled.

makspll:
Oh yes I think tealr also updated the mlua version which had some big changes to the way `FromLua` worked

LaserWitch:
Superseded for now by #90

PR #93, opened 2023-12-17
Bevy 0.12 migration

LaserWitch:
This effort has reached a point where examples are buildable but the asset handling doesn't actually work right yet.
Bevy 0.12 had breaking changes in parts of assets and reflection that are pretty intimately intwined with the API code. I've made a number of edits in here that were just to get things building, and not always well researched correct changes. But it does actually build against bevy 0.12 so I figure it's worth making a draft.

~~Includes the changes from #92 currently.~~


LaserWitch:
The failed checks remind me to say that I'm currently focusing on lua as I actually know how to work with it's test cases. If I get everything working there, maybe I'll try to get the rhai stuff at least building too.

makspll:
Thanks @LaserWitch! That looks good, just added the Rhai bits + commited minor changes

Issue #97, opened 2024-01-18
lua globals are not registered until callback function.


ConnorBP:
Globals are not being defined on my lua script until partway into it's execution (when a callback is called). Is this how it is supposed to go? and can it be changed? or is this a bug?

For example, in this modified version of console_integration.lua the `print_to_console` function is not defined in the beginning of the script. It does work however in the on_update function. This makes it hard to reliably override the print function since it only sometimes applies.
```lua
local a = 0

print("testing print")
print_to_console("testing print2")

function on_update()

    if a % 100 == 0 then
        -- print() is defined in lua_system.rs
        -- by the api provider
        print_to_console(string.format("%d, entity index:%d", a, entity:index()))
    end

    a = a + 1
end
```

ConnorBP:
Another thing:

```rs
pub fn forward_script_err_to_console(
    mut r: EventReader<ScriptErrorEvent>,
    mut w: EventWriter<PrintConsoleLine>,
) {
    for e in r.iter() {
        w.send(PrintConsoleLine {
            line: format!("ERROR:{}", e.error).into(),
        });
    }
}
```
is not forwarding errors when a script fails to load:
```
2024-01-18T21:55:17.170874Z  WARN bevy_mod_scripting_core::hosts: Error in loading script console_example.lua:
Failed to load script asset for `console_example.lua` runtime error: [string "console_example.lua"]:4: attempt to call global 'print_test' (a nil value)
stack traceback:
        [C]: in function 'print_test'
        [string "console_example.lua"]:4: in main chunk
```

ConnorBP:
> Another thing:
> 
> ```rust
> pub fn forward_script_err_to_console(
>     mut r: EventReader<ScriptErrorEvent>,
>     mut w: EventWriter<PrintConsoleLine>,
> ) {
>     for e in r.iter() {
>         w.send(PrintConsoleLine {
>             line: format!("ERROR:{}", e.error).into(),
>         });
>     }
> }
> ```
> 
> is not forwarding errors when a script fails to load:
> 
> ```
> 2024-01-18T21:55:17.170874Z  WARN bevy_mod_scripting_core::hosts: Error in loading script console_example.lua:
> Failed to load script asset for `console_example.lua` runtime error: [string "console_example.lua"]:4: attempt to call global 'print_test' (a nil value)
> stack traceback:
>         [C]: in function 'print_test'
>         [string "console_example.lua"]:4: in main chunk
> ```

[...]

PR #98, opened 2024-01-19
Apply LUA api providers before loading script.

ConnorBP:
Fixes issue where api is not available during script initialization. https://github.com/makspll/bevy_mod_scripting/issues/97

PR is not yet complete. There is still a `callback error` caused by: error converting Lua nil to userdata to investigate which happens when the api function is called.


ConnorBP:
Found what is causing the NIL error. It happened upon fetching the bevy "world" from globals. It is not set on script load execution, only after event execution is fired. This makes it so that any API function which accesses world state cannot be used during script load. To fix this requires a refactor involving changing the `script_add_synchronizer` into an `exclusive system` with a `SystemState`

ConnorBP:
Update: the other issues have their own concerns that should be in another PR. But the api registration issue is solved by this PR and should be considered Complete.

I will open a different PR for potential api refactoring for connecting callback functions with event signals.

languages/bevy_mod_scripting_lua/src/lib.rs
```diff
@@ -113,17 +113,24 @@ impl<A: LuaArg> ScriptHost for LuaScriptHost<A> {
         #[cfg(not(feature = "unsafe_lua_modules"))]
         let lua = Lua::new();
 
-        lua.load(script)
+        // init lua api before loading script
+        let mut lua = Mutex::new(lua);
```
makspll, line 117: Could we wrap in the mutex at the end instead?

makspll (changes requested):
Looks good, just one small nit

ConnorBP:
> Looks good, just one small nit

`providers.attach_all()' requires a mutex as it's argument.

makspll:
Sounds good, let's just fix CI and happy to merge

makspll (approved):

PR #99, opened 2024-01-20
Rune support

baby-steve:
This adds a `ScriptHost` implementation for the Rune programming language. It's essentially a copy and paste of Lua and Rhai's  hosts, nothing super different or anything.

The two examples are mostly there for testing. I can remove them, change them or add new ones if you'd like.


languages/bevy_mod_scripting_rune/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+                .setup_runtime_all(world.clone(), &script_data, ctx)
+                .expect("Could not setup script runtime");
+
+            for event in events {
+                if !event.recipients().is_recipient(&script_data) {
+                    continue;
+                }
+
+                // TODO: should we store `Vm` in a script context?
+                let mut vm = Vm::new(ctx.runtime_context.clone(), ctx.unit.clone());
```
makspll, line 228: I am not familiar with Rune, but this seems like what should be stored as the script context, copying the entire context, and creating a new stack is definitely not the way to do this :)

makspll (changes requested):
Major:
- we need to store the VM as the script context I think, or at least have the VM cached
Minor:
- can we add the rune feature to the CI scripts?

languages/bevy_mod_scripting_rune/src/lib.rs
baby-steve: Yeah, we should be storing a rune `Vm` per script as the script context. The only problem is that `Vm` doesn't implement `Send` or `Sync` which `ScriptHost::ScriptContext` requires. For whatever reason, sticking the vm in a `Mutex` doesn't work. 

Caching it could work. We could store `Vm`s in a non send resource or something. 

Do script contexts ever get accessed outside of the main thread or a single thread? From what I can tell, they don't. If they aren't, then I could wrap `Vm` in a new type and (unsafely) implement `Send` and `Sync` for it. So long as `handle_events` is the only system accessing them, this should be safe. 

(Also, the runtime context is inside an `Arc` so cloning it is cheap. Same with unit).

baby-steve: Disregard my previous comment. Turns out that you can reuse a `Vm` for different units/contexts, which means we could store a single `Vm` in a non-send resource and reuse it for all script executions (I've never had to explore that corner of rune's API until now ¯\\\_(ツ)_/¯). I'm going to go ahead and implement it that way.

makspll (approved):
Looks good! Thanks for this

PR #103, opened 2024-02-04
Script Management Exclusive System Refactor (NEEDS REVIEW)

ConnorBP:
Tentative initial work for exclusive system refactor. Still in todo:

- [ ] quality check some of the changes (spent most of the time fighting the borrow checker). 
- [ ] make sure the new interface is correctly added to each script host implementation
- [x] figure out why even with this, trying to get  world from an attach api bound function is returning nil
- [ ] implement usage of world access to rune and rhai (maybe this should be its own PR)
- [ ] the script reload system makes a clone of all loaded script handles due to lifetime reasons. Lessening this would be ideal.
- [ ] update relevant documentation about script load system
- [ ] maybe increment semvar?

example of what is trying to be fixed
```rs
fn attach_api(&mut self, ctx: &mut Self::APITarget) -> Result<(), ScriptError> {
        // callbacks can receive any `ToLuaMulti` arguments, here '()' and
        // return any `FromLuaMulti` arguments, here a `usize`
        // check the Rlua documentation for more details

        println!("ATTACHED API");

        let ctx = ctx.get_mut().unwrap();

        ctx.globals()
            .set(
                "print2",
                ctx.create_function(|ctx, msg: String| {
                    println!("RUNNING PRINT COMMAND");
                    // retrieve the world pointer
                    let world = match (ctx.get_world()) { // <---------------------------------- THIS FAILS when called during script load
                        Ok(world) => world,
                        Err(e) => {
                            error!("print() failed to get world from ctx: {e}");
                            return Ok(());
                        }
                    };
                    println!("GOT WORLD");
                    let mut world = world.write();
                    println!("GOT WRITE ACCESS TO WORLD");

                    let mut events: Mut<Events<PrintConsoleLine>> =
                        world.get_resource_mut().unwrap();
                    println!("GOT EVENT WRITER");
                    events.send(PrintConsoleLine { line: msg.into() });
                    println!("SENT EVENT");

                    // return something
                    Ok(())
                })
                .map_err(|e| {
                    println!("print was called and errored. {e}");
                    ScriptError::new_other(e)
                })?,
            )
            .map_err(|e| {
                println!("failed to attach print command. {e}");
                ScriptError::new_other(e)
            })?;

        // equivalent to ctx.globals().set() but for multiple items
        tealr::mlu::set_global_env(Export, ctx).map_err(ScriptError::new_other)?;

        Ok(())
    }
```

# List of changes to api from this PR:

- Made script load and reload api and systems intake a global varialbe
- replaced `event_writer: &mut EventWriter<ScriptLoaded>,` argument to reload and insert new functions with a resource_scope on the world pointer.
- Both api bindings and "world" global are now available during initial script load execution and api initialization
- Script handles are now cloneable and so are script collections
- more script errors are now passed to the event writer (such as script load failure)
- added CachedScriptLoadState system state resource for event writers and queries used by the exclusive system script loading systems

### DIFF
--- a/bevy_mod_scripting_core/src/event.rs
+++ b/bevy_mod_scripting_core/src/event.rs
@@ -1,6 +1,10 @@
+use bevy::log::error;
 use bevy::prelude::Event;
 
 use crate::{error::ScriptError, hosts::Recipients};
+use crate::hosts::ScriptHost;
+use crate::systems::CachedScriptState;
+use crate::world::WorldPointer;
 
 /// An error coming from a script
 #[derive(Debug, Event)]
@@ -19,4 +23,20 @@ pub struct ScriptLoaded {
 pub trait ScriptEvent: Send + Sync + Clone + Event + 'static {
     /// Retrieves the recipient scripts for this event
     fn recipients(&self) -> &Recipients;
+}
+
+pub fn write_error_event_with_world<H: ScriptHost>(world: WorldPointer, script_name: String, error_text: String) {
+    let mut world = world.write();
+    let mut state: CachedScriptState<H> = world.remove_resource().unwrap();
+
+    let (_, mut error_wrt, _) = state.event_state.get_mut(&mut world);
+
+    let error = ScriptError::RuntimeError {
+        script: script_name,
+        msg: error_text,
+    };
+
+    error!("{}", error);
+    error_wrt.send(ScriptErrorEvent { error });
+    world.insert_resource(state);
 }
--- a/bevy_mod_scripting_core/src/event.rs
+++ b/bevy_mod_scripting_core/src/event.rs
@@ -1,10 +1,10 @@
 use bevy::log::error;
 use bevy::prelude::Event;
 
-use crate::{error::ScriptError, hosts::Recipients};
 use crate::hosts::ScriptHost;
 use crate::systems::CachedScriptState;
 use crate::world::WorldPointer;
+use crate::{error::ScriptError, hosts::Recipients};
 
 /// An error coming from a script
 #[derive(Debug, Event)]
@@ -25,7 +25,11 @@ pub trait ScriptEvent: Send + Sync + Clone + Event + 'static {
     fn recipients(&self) -> &Recipients;
 }
 
-pub fn write_error_event_with_world<H: ScriptHost>(world: WorldPointer, script_name: String, error_text: String) {
+pub fn write_error_event_with_world<H: ScriptHost>(
+    world: WorldPointer,
+    script_name: String,
+    error_text: String,
+) {
     let mut world = world.write();
     let mut state: CachedScriptState<H> = world.remove_resource().unwrap();
 
--- a/bevy_mod_scripting_core/src/hosts.rs
+++ b/bevy_mod_scripting_core/src/hosts.rs
@@ -1,11 +1,13 @@
 //! All script host related stuff
 use bevy::{asset::Asset, ecs::schedule::ScheduleLabel, prelude::*};
+use std::ops::DerefMut;
 use std::{
     collections::HashMap,
     iter::once,
     sync::atomic::{AtomicU32, Ordering},
 };
 
+use crate::world::WorldPointerGuard;
 use crate::{
     asset::CodeAsset,
     docs::DocFragment,
@@ -73,6 +75,7 @@ pub trait ScriptHost: Send + Sync + 'static + Default + Resource {
     /// to send useful errors.
     fn load_script(
         &mut self,
+        world_ptr: WorldPointer,
         script: &[u8],
         script_data: &ScriptData,
         providers: &mut APIProviders<Self>,
@@ -113,11 +116,24 @@ pub trait ScriptHost: Send + Sync + 'static + Default + Resource {
         };
 
         let mut providers: APIProviders<Self> = world.remove_resource().unwrap();
-        let mut ctx = self.load_script(script, &fd, &mut providers).unwrap();
+        // safety:
+        // - we have &mut World access
+        // - we do not use the original reference again anywhere in this function after this was created
+        let world = unsafe { WorldPointerGuard::new(world) };
+        let mut ctx = self
+            .load_script(world.clone(), script, &fd, &mut providers)
+            .unwrap();
         self.setup_script(&fd, &mut ctx, &mut providers)?;
         let events = [event; 1];
 
-        self.handle_events(world, &events, once((fd, &mut ctx)), &mut providers);
+        let mut world = world.write();
+
+        self.handle_events(
+            world.deref_mut(),
+            &events,
+            once((fd, &mut ctx)),
+            &mut providers,
+        );
 
         world.insert_resource(providers);
 
@@ -352,12 +368,12 @@ impl<T: Asset> Script<T> {
     /// reloads the script by deleting the old context and inserting a new one
     /// if the script context never existed, it will after this call.
     pub(crate) fn reload_script<H: ScriptHost>(
+        world: &mut World,
         host: &mut H,
         script: &Script<H::ScriptAsset>,
         script_assets: &Assets<H::ScriptAsset>,
         providers: &mut APIProviders<H>,
         contexts: &mut ScriptContexts<H::ScriptContext>,
-        event_writer: &mut EventWriter<ScriptLoaded>,
     ) {
         debug!("reloading script {}", script.id);
 
@@ -367,13 +383,13 @@ impl<T: Asset> Script<T> {
             contexts.remove_context(script.id());
             // insert new re-loaded context
             Self::insert_new_script_context::<H>(
+                world,
                 host,
                 script,
                 entity,
                 script_assets,
                 providers,
                 contexts,
-                event_writer,
             );
         } else {
             // remove old context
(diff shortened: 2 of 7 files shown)


ConnorBP:
The console integration examples print command still fails to get LuaWorld from "world" on script load for some reason here. Not quite sure why....

I'm gonna come back to it with fresh eyes and maybe realize that I forgot to call a function somewhere, but please do let me know if you spot it.

ConnorBP:
I should mention, one thing the changes so far have made possible which was not possible previously is this use case:

```rs
#[derive(Default)]
struct Export;
impl tealr::mlu::ExportInstances for Export {
    fn add_instances<'lua, T: tealr::mlu::InstanceCollector<'lua>>(
        self,
        instance_collector: &mut T,
    ) -> mlua::Result<()> {
        instance_collector
            // .document_instance("The root collection of API modules making up the Lychee API")
            // .add_instance("lychee", |ctx| {
            //     Ok(ctx.create_table_from(vec![("test", LycheeAPIModule)])?)
            // })?
            .document_instance("Utility functions for interacting with the game memory directly.")
            .add_instance("memory", |ctx| {
                if let Some(processr) = ctx
                    .get_world()
                    .map_err(|e| {
                        mlua::Error::RuntimeError(format!(
                            "Failed getting world in tealr lua api init: {e}"
                        ))
                    })?
                    .read()
                    .get_resource::<ProcessR<IntoProcessInstance<CBox<c_void>, CArc<c_void>>>>()
                {
                    Ok(MemoryAPIModule {
                        process: processr.0.clone(),
                    })
                } else {
                    return Err(mlua::Error::RuntimeError(
                        "failed to get memflow process resource".to_string(),
                    ));
                }
            })
            .map_err(|e| mlua::Error::RuntimeError(format!("{e}")))?;

        Ok(())
    }
}
```

Before the changes in this PR the globals variable was not set so it was not possible to initialize modules content in the instance collector with any meaningful data.

The fact that this is working now means that the globals are in-fact set, so I am not quite sure why they are still inaccessible during script load execution. Something weird with the stack is going on I think?

ConnorBP:
> Found the problem I believe. Its a cyclic dependency. The world cannot be injected the first time and fails because its type has no yet been defined by tealr in the add_instances. But if you call attach api first then the world passed to it is not yet set.
> 
> ```rust
> set("world", crate::lua::bevy::LuaWorld::new(world_ptr))
> ```
> 
> fails if
> 
> ```rust
>  instances.add_instance(
>     "world",
>     crate::lua::util::DummyTypeName::<crate::lua::bevy::LuaWorld>::new,
> )?;
> ```
> 
> was not called first.
> 
> Not sure what the best solution is. The quickest fix would be to call the initialization twice, but some way of splitting off the dependent sections into a third seperate stage would be ideal I think.
> 
> * Register Types by attaching bevy_api
> * then setup runtime which causes world to be set
> * then attach any apis dependent on world access

I was wrong. I don't think this was relevant at all. I now suspect it is because ` contexts.insert_context(script_data.to_owned(), Some(lua));` is called *after* the load script function.

ConnorBP:
I have been debugging this for days. I am completely at a loss as to why this function fails when called in a rust callback function that is bound to lua:
```rs
impl GetWorld for Lua {
    type Error = mlua::Error;
    fn get_world(&self) -> Result<WorldPointer, Self::Error> {
        self.globals().get::<_, LuaWorld>("world").map(Into::into)
    }
}
```
`.get::<_, LuaWorld>("world")` here returns a nil. The craziest thing is that the world global is accessible just fine from the tealr `add_instances()` function which is called shortly before this one. Either something is funky with the way the lua callbacks are being created, or world is being popped without being put back. I am very close to giving up but I might be close........ I have been close for days. :'(

makspll:
So i've had a quick look, Looks like the issues begin with this code:

```
        // {
        //     providers.attach_all(&mut lua)?;
        // }
```

commenting this out and modifying the console example script to:

image

Gives me this output:

image

so somehow the API attachment step resets the globals maybe?

makspll:
Okay, here is the easy solution then, setting up the script runtime is very light, it's just setting up a world pointer, let's do it twice, once before attaching hte API so it's available there, then afterwards so it's available when the script loads, this shouldn't mess with docs generation and won't drag performance too much (plus only once per script). Otherwise we'd have to figure out another way of documenting these things but the tealr part of the system needs an overhaul later anyway so that can be addressed at the same time.

ConnorBP:
> Okay, here is the easy solution then, setting up the script runtime is very light, it's just setting up a world pointer, let's do it twice, once before attaching hte API so it's available there, then afterwards so it's available when the script loads, this shouldn't mess with docs generation and won't drag performance too much (plus only once per script). Otherwise we'd have to figure out another way of documenting these things but the tealr part of the system needs an overhaul later anyway so that can be addressed at the same time.
> 
> EDIT: Yes doing this fixed the problem

Oh amazing! Nice work, thanks for the assist!

makspll:
Hey @ConnorBP sorry for not reviewing in a while, I am doing some serious re-factoring and I have this issue in mind. I think soon this use case will be supported through a much more modular structure!

ConnorBP:
> Hey @ConnorBP sorry for not reviewing in a while, I am doing some serious re-factoring and I have this issue in mind. I think soon this use case will be supported through a much more modular structure!

Excellent! :) should be much easier on the latest bevy as well. The branch for this PR works pretty well if you wanna try it out

shanecelis:
If this PR makes `world` available at script load time, I'm all about that! Any chance it could be resurrected? I see it's a little dated now. If it was rebased and working with main, would it be likely to be merged?

makspll:
Hi @shanecelis, if I remember correctly the PR needed some more work before merging (testing, and I think porting this to the other scripting langs), if the demand is high I don't mind merging this work assuming somebody else does this work. However worth keeping in mind is these features will be supported in the big upcoming refactor over at: https://github.com/makspll/bevy_mod_scripting/pull/112 (but I cannot tell you when that will be ready)

ConnorBP:
> ), if the demand is high I don't mind merging this work assuming somebody else does this work. However worth keeping in mind is these features will be supported in the big upcoming

Hello, sorry for the delay. It was mostly done. I think I moved on to other stuff for a while because I wasn't sure if you would like the design decisions I took in this refactor. If your newest PR isn't incompatible I can definitely finish this one up.

makspll:
> Hello, sorry for the delay. It was mostly done. I think I moved on to other stuff for a while because I wasn't sure if you would like the design decisions I took in this refactor. If your newest PR isn't incompatible I can definitely finish this one up.

No problem at all! This is more on me for working on this crate at yearly intervals versus small changes over time!


>  Both api bindings and "world" global are now available during initial script load execution and api initialization

Yes! I am currently writing a bunch of integration tests, which act exactly like normal scripts (get loaded then `on_test` gets called as a callback), you can see this one: https://github.com/makspll/bevy_mod_scripting/blob/feature/bevy-system-refactor/crates/languages/bevy_mod_scripting_lua/tests/data/api_availability/api_available_on_script_load.lua and this one https://github.com/makspll/bevy_mod_scripting/blob/feature/bevy-system-refactor/crates/languages/bevy_mod_scripting_lua/tests/data/api_availability/api_available_on_callback.lua both pass!

> Made script load and reload api and systems intake a global variable

Yes, although now the concept has been slightly abstracted: https://github.com/makspll/bevy_mod_scripting/blob/01dcf8ee9fdba0c0c06f963aee2b833ab8f59f61/crates/bevy_mod_scripting_core/src/context.rs#L90

> replaced event_writer: &mut EventWriter, argument to reload and insert new functions with a resource_scope on the world pointer.

I think everything now pretty much gets exclusive world access

> Script handles are now cloneable and so are script collections
> more script errors are now passed to the event writer (such as script load failure)

Hmm no, but I can make it so!

> added CachedScriptLoadState system state resource for event writers and queries used by the exclusive system script loading systems

Less relevant now due to bevy improvements to exclusive systems I believe.

ConnorBP:
> > Hello, sorry for the delay. It was mostly done. I think I moved on to other stuff for a while because I wasn't sure if you would like the design decisions I took in this refactor. If your newest PR isn't incompatible I can definitely finish this one up.
> 
> 
> 
> No problem at all! This is more on me for working on this crate at yearly intervals versus small changes over time!
> 
> 
> 
> 
> 
> >  Both api bindings and "world" global are now available during initial script load execution and api initialization
> 
> 
> 
> Yes! I am currently writing a bunch of integration tests, which act exactly like normal scripts (get loaded then `on_test` gets called as a callback), you can see this one: https://github.com/makspll/bevy_mod_scripting/blob/feature/bevy-system-refactor/crates/languages/bevy_mod_scripting_lua/tests/data/api_availability/api_available_on_script_load.lua and this one https://github.com/makspll/bevy_mod_scripting/blob/feature/bevy-system-refactor/crates/languages/bevy_mod_scripting_lua/tests/data/api_availability/api_available_on_callback.lua both pass!
> 
> 
> 
> > Made script load and reload api and systems intake a global variable
> 
> 
> 
> Yes, although now the concept has been slightly abstracted: https://github.com/makspll/bevy_mod_scripting/blob/01dcf8ee9fdba0c0c06f963aee2b833ab8f59f61/crates/bevy_mod_scripting_core/src/context.rs#L90
> 
> 
> 
> > replaced event_writer: &mut EventWriter, argument to reload and insert new functions with a resource_scope on the world pointer.
> 
> 
> 
> I think everything now pretty much gets exclusive world access
> 
> 
> 
> > Script handles are now cloneable and so are script collections
> 
> > more script errors are now passed to the event writer (such as script load failure)
> 
> 
> 
> Hmm no, but I can make it so!
> 
> 
> 
> > added CachedScriptLoadState system state resource for event writers and queries used by the exclusive system script loading systems
> 
> 
> 
> Less relevant now due to bevy improvements to exclusive systems I believe.
> 
> 
> 
> 
> 
> 

Oh awesome! Okay this PR is out-moded then 😀 I'll close it in favour of your new branch

PR #114, opened 2024-04-08
Make generated wrappers publically accessible

makspll:


LaserWitch (approved):
This appears to work and fix the (informally reported) issue I was having migrating to 0.6. Might be worth a 0.6.1 release?

PR #118, opened 2024-04-26
Dynamic Queries

### LINKED ISSUE #20: Dynamic queries
Allow scripts querying for entities with different components.
- [ ] Write (Owned only) Lua wrappers around `bevy_ecs_dynamic`
- [ ] Add them to the APIProvider

Mistrustfully:
Fixes #20 
Adds API methods for dynamic queries, using the new QueryBuilder added in Bevy 0.13.
Both Lua and Rhai have implementations as well as types for Teal.

Lua:
```lua
local ComponentOne = world:get_type_by_name("ComponentOne")
local ComponentTwo = world:get_type_by_name("ComponentTwo")
local ComponentThree = world:get_type_by_name("ComponentThree")
local ComponentFour = world:get_type_by_name("ComponentFour")

-- `world:query` accepts any number of `LuaTypeRegistration`
-- returns a `LuaQueryBuilder` which can be used to further filter the query with the `:with` and `:without` methods
-- `LuaQueryBuilder` has an `:iter` method which resolves the query and returns an iterator over the entities and components

for entity, componentOne, componentTwo in world:query(ComponentOne, ComponentTwo):with(ComponentThree):without(ComponentFour):iter() do
    print(entity, componentOne, componentTwo)
end
```

Rhai:
```rs
let component_one = world.get_type_by_name("ComponentOne");
let component_two = world.get_type_by_name("ComponentTwo");
let component_three = world.get_type_by_name("ComponentThree");
let component_four = world.get_type_by_name("ComponentFour");

// `world.query` accepts an array of `ScriptTypeRegistration`
// returns a `ScriptQueryBuilder` which can be used to further filter the query with the `.with` and `.without` methods
// Iterating over `ScriptQueryBuilder` will resolve the query

for results in world.query([ ComponentOne, ComponentTwo ]).with([ ComponentThree ]).without([ ComponentFour ]) {
    // `results` is an object map, keys being the type name of the component, and the values being the entities' component
    // `results` also has a `Entity` key, being the entity the query matched against
    print(results.Entity);
    print(results.ComponentOne);
    print(results.ComponentTwo);
}
```

### DIFF
--- a/crates/bevy_script_api/src/common/bevy/mod.rs
+++ b/crates/bevy_script_api/src/common/bevy/mod.rs
@@ -1,12 +1,12 @@
-use std::{
-    ops::{Deref, DerefMut},
-    sync::Arc,
-};
-
 use crate::ReflectReference;
 /// Common functionality for all script hosts
 use bevy::{
-    ecs::system::Command,
+    ecs::{
+        component::ComponentId,
+        query::QueryBuilder,
+        system::Command,
+        world::{EntityRef, World},
+    },
     prelude::{
         AppTypeRegistry, BuildWorldChildren, Children, DespawnChildrenRecursive, DespawnRecursive,
         Entity, Parent, ReflectComponent, ReflectDefault, ReflectResource,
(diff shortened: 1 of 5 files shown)


makspll:
Very solid work, Much appreciated! It's also very nice that it mostly fit without much trouble into the existing framework!

PR #127, opened 2024-09-07
Migrate to bevy 0.14

### LINKED ISSUE #124: Support for bevy 0.14
I tried doing this myself. At first it didn't seem like it would be much work. However, once I got to the autogenerated code, I got lost. Is there any documentation or guidance for how you go about updating your bevy version? I'd be happy to attempt a PR.

makspll:
Will fix #124


makspll:
I am getting stack overflows on running codegen, which is strange, I would expect proper compilation failures

makspll:
Oh it looks to be in the active feature computation stage which does make sense as the crate sets some features for bevy which might be out of date

makspll:
There seems to be a dependency loop in bevy, or at least in the way the dependncies are processed

makspll:
It seems to have been an issue in the way we computed dependencies, since dev dependencies were included and that lead to inifnite cycles between a crate and itself (bevy_math). 

It seems we get compilation errors on bevy from new rust compiler features, so I do need to upgrade the codegen nightly version

makspll:
This seems to be working now, examples are working fine

shanecelis:
I'm positively giddy about this!

shanecelis:
Got a chance to kick the tires and successfully updated my project to use the main branch and bevy 0.14. No issues on my end.

PR #130, opened 2024-10-20
Documentation generation hotfixes

ischeinkman:
Hotfix for 2 bugs in the teal environment generation algorithm: 

* Fixed fields not being deduped correctly because `Vec::dedup` only dedupes *consecutive* duplicate items. This was done by sorting the fields list beforehand. 
    * Also deduped static fields while I was at it, though I'm not sure if those are used? 
* Fixed the algorithm not properly escaping record fields/functions/methods/etc that share the name of an existing Lua/Teal keyword.  This was coming up since the `Uuid` record has a function called `nil` for making an all-zero uuid. 

Note that this still doesn't completely fix Teal document generation when using the default Bevy APIs since they're still missing a definition for `ReflectedValue`, but I'll likely need some guidance on that and will split that off to a separate PR.


gallexme:
Tested it works great,

for the  ReflectedValue i had to add
  .process_type::<crate::script_ref::ReflectedValue>()
crates/bevy_script_api/src/core_providers.rs  get_doc_fragment

PR #183, opened 2024-12-31
feat: Complete Rewrite

makspll:
Todo:
- [x] Global namespace support
- [x] Removal/Overwriting of script functions
- [x] Re-check soundness
- [x] Increase coverage in BMS core
- [x] Add insert component/resource methods
- [x] Decouple code gen away from lua
- [x] Get rid of panicky errors on access violations and in general reduce `unwrap` usage
- [x] ~~Get rid of script allocator in favour of an actual allocator~~ not possible at the moment
- [x] RHAI
- [x] Optional args
- [x] Documentation registration scaffolding
- [x] Utilities for getters/setters
- [x] HashMap support 
- [x] final docs adjustments


makspll:
There is potential to make the book both a reference and source of integration tests for the various languages

makspll:
Now released as 0.9.0-alpha.2

shanecelis:
# Important: Keep Going!

This is looking really good! I went through the book. It looks a lot simpler for the user and more controllable too. I think I'm going to try to build with the alpha.5 for my project [nano-9](https://github.com/shanecelis/nano-9) soon.

# Unimportant: A Question

One question that's not clear to me is, what is shared between scripts? Do they each have their own context? Is `Runtime` the facilitator for that?

One thing I'd like to be able to do is load some Lua code for every script. Playing with the example, I got this to work:

```rust
    let mut plugin = LuaScriptingPlugin::default();
    plugin.scripting_plugin.add_context_initializer(|script_id: &str, context: &mut bevy_mod_scripting::lua::mlua::Lua| {
        let _ = context.load(r#"
my_global = "BLAH"
"#).exec()?;
        Ok(())
    });
```

Is this what you'd suggest for loading static Lua code? Your documentation suggested this was good when you'd had dynamic needs.

I tried simply loading a "script.lua" but the two namespaces appear to be isolated.

I apologize for adding some tech support questions in this PR. I'm not blocked; I don't need any answers. Just thought I'd share what my sticking point is from a 0.8 user.

makspll:
Hi! All good!

Yes if you have something you'd like to inject into every script, I'd reccomend using context initializers!

As for sharing contexts, I've made it possible to change context loading settings, so that all scripts actually do end up living in a single Lua context.

I'll add a bit to the docs about that

shanecelis:
I got [it](https://github.com/shanecelis/nano-9/tree/bms-alpha) compiling! One snag I haven't overcome: How can I pass this `N9Entity` as a `ScriptValue`?

```rust
#[derive(Debug, Clone, Copy, Reflect)]
pub enum DropPolicy {
    Nothing,
    Despawn,
}

#[derive(Clone, Reflect)]
pub struct N9Entity {
    pub entity: Entity,
    pub drop: DropPolicy,
}
```

I'd been passing an `Arc<Mutex<N9Entity>>` before, but if you could show me or suggest how to pass an `N9Entity` as a `ScriptValue`, I can make do I think.

gallexme:
> I got [it](https://github.com/shanecelis/nano-9/tree/bms-alpha) compiling! One snag I haven't overcome: How can I pass this `N9Entity` as a `ScriptValue`?
> 
> ```rust
> #[derive(Debug, Clone, Copy, Reflect)]
> pub enum DropPolicy {
>     Nothing,
>     Despawn,
> }
> 
> #[derive(Clone, Reflect)]
> pub struct N9Entity {
>     pub entity: Entity,
>     pub drop: DropPolicy,
> }
> ```
> 
> I'd been passing an `Arc<Mutex<N9Entity>>` before, but if you could show me or suggest how to pass an `N9Entity` as a `ScriptValue`, I can make do I think.

Checkout  `ReflectReference::new_allocated(N9Entity,allocator)`
Using the allocator u can get from world 
https://github.com/makspll/bevy_mod_scripting/blob/f71561f3cfbc9e41a53577422aa1ed0bc81feeca/crates/script_integration_test_harness/src/test_functions.rs


It's a bit fluid atm amd still changing

rdrpenguin04:
Three months ago, I was looking for a crate like this.
Three weeks ago, I discovered this crate.
Three hours ago, I decided to experiment with it.
Three minutes ago, I saw rumblings of a rewrite and found the Battle Plan issue.
Three seconds ago, I found this PR.
And in three months, I'll be as committed as I can be to helping develop this.

Thank you for all you've done and are continuing to do! I'm so excited for the new version to land, and I'm excited for what comes next!

Do I have anything of actual substance to comment here? ... No, it's 5:13 AM, and I should've gone to bed at least four hours ago. But I wanted you to know that you're doing something amazing in more ways than just a star can express.

makspll:
Appreciate all the love! The rewrite is now in a pretty good place, I don't expect many big changes from now :D 
 
> I got [it](https://github.com/shanecelis/nano-9/tree/bms-alpha) compiling! One snag I haven't overcome: How can I pass this N9Entity as a ScriptValue?

Indeed you need the `ReflectAllocator`, I was experimenting with making a full blown allocator to get rid of this necessity, but that would require:
- the `allocator_api` to get stabilised in rustc
- reflection to expose layout information about types

The reason the allocator is needed, is because I need guarantees that things won't get deallocated untill after the script handlers are finished

I might be able make the allocator a static at least which could make things easier to use, but for now I might keep it as is

shanecelis:
The conversion goes well. Many things are working. And having spent more time with it, this BMS rewrite is HUGE improvement. I do seem to have hit one issue though. [One of my code paths](https://github.com/shanecelis/nano-9/blob/2ec91252cfff004bf4b3ae1dde50532accb52e2e/src/pico8/cart.rs#L310) constructs a `ScriptAsset` from bytes (the Lua content is embedded in another file). I imagine if I did `asset_server.load()` it would work. Any advice?

Also I had to derive `Reflect` on `GlobalNamespace` to get `NamespaceBuilder::<GlobalNamespace>::new(&mut world)` to work. Not sure if that's how it should be, but that was my workaround.

makspll:
Great to hear! 

When it comes to this nested asset thing, ill need to investigate a bit more,

But there is a possibility to load non asset backed scripts by issuing the create and delete commands yourself, the handle to the asset there is optional.

If you get your pipeline to emmit normal asset events for scripts everything will work automatically, so thats what I'd look out for.

As for rhe Global namespace, there is a `new_unregistered` method on the builder which doesn't require reflect on the type!

rdrpenguin04:
Would it be possible to reflect operator overloads into Lua (`Add` to `__add`, etc)? If there isn't a path to this yet, I may do some digging to get that machinery in; I have a specific use separate from just vector math.

makspll:
> Would it be possible to reflect operator overloads into Lua (`Add` to `__add`, etc)? If there isn't a path to this yet, I may do some digging to get that machinery in; I have a specific use separate from just vector math.

There is limited support for it in the codegen, as in it's mapped via macros for glam types only, but it should be possible to  modify this logic to work on any type. I wasn't planning on doing this myself but would be open to PR's here!

shanecelis:
I found my issue. My cart's lua code had this path: "cart.p8#lua" and that wasn't recognized as a `Language::Lua`. Got it fixed up with `ScriptAssetSettings`.

One new issue is, how do you handle optional arguments? I was hoping I could use `Option` like so:

```rust
        NamespaceBuilder::<GlobalNamespace>::new_unregistered(&mut world)
            .register(
                "btn",
                |ctx: FunctionCallContext, b: Option<u8>| {
                    with_pico8(&ctx, |pico8| pico8.btn(b))
                },
            );
```

But I get, "Error in function btn : Error in function call: expected 1 arguments but received 0." On that note, how do you handle varargs?

makspll:
> I found my issue. My cart's lua code had this path: "cart.p8#lua" and that wasn't recognized as a `Language::Lua`. Got it fixed up with `ScriptAssetSettings`.

Nice! I will add this to the docs as well

> One new issue is, how do you handle optional arguments? I was hoping I could use `Option` like so:
> 
> ```rust
>         NamespaceBuilder::<GlobalNamespace>::new_unregistered(&mut world)
>             .register(
>                 "btn",
>                 |ctx: FunctionCallContext, b: Option<u8>| {
>                     with_pico8(&ctx, |pico8| pico8.btn(b))
>                 },
>             );
> ```
> 
> But I get, "Error in function btn : Error in function call: expected 1 arguments but received 0." On that note, how do you handle varargs?

Ah yes, this will expect a unit argument, i.e. Nil instead of a lack of arguments. Hmm, i could make the calling mechanism recognize "optional" arguments. 

As for varargs, there isn't support for this yet, I think I could however add a variant to the script value which marshalls appropriately 🤔

shanecelis:
Thank you! 

I puzzled this one out myself. Haven't figured out how to exercise it. Looks like `FunctionCallContext` must come first. 

```rust
    NamespaceBuilder::<N9Entity>::new(app.world_mut())
        .register("name", |ctx: FunctionCallContext, this: Val<N9Entity>| {
            let world = ctx.world()?;
            // let id: Entity = Entity::PLACEHOLDER; // How do I get the N9Entity's entity field? Oh, I see.
            world.with_component(this.entity, |name: Option<&Name>| {
                name.map(|s| s.as_str().to_owned())
            })
        })
```
Is this going to do what I want? My hope would be in Lua that I could do `some_n9entity:name()` and get nil or its name if any.

Is this what we do now instead of implementing `UserData`?

shanecelis:
Oh, one thing I thought I'd pass on. You've got some functions with this signature:
```rust
pub fn register_script_functions(app: &mut App) -> &mut App;
```
Perhaps they might be good candidates for the function as `Plugin`:

```rust
pub fn register_script_functions(app: &mut App) {
    NamespaceBuilder::<World>::new_unregistered(app.world_mut()).register("info", |s: String| {
        bevy::log::info!(s);
    });
}
/// ...
app.add_plugins(register_script_functions);
```

makspll:
> Thank you! 
> 
> I puzzled this one out myself. Haven't figured out how to exercise it. Looks like `FunctionCallContext` must come first. 
> 
> ```rust
>     NamespaceBuilder::<N9Entity>::new(app.world_mut())
>         .register("name", |ctx: FunctionCallContext, this: Val<N9Entity>| {
>             let world = ctx.world()?;
>             // let id: Entity = Entity::PLACEHOLDER; // How do I get the N9Entity's entity field? Oh, I see.
>             world.with_component(this.entity, |name: Option<&Name>| {
>                 name.map(|s| s.as_str().to_owned())
>             })
>         })
> ```
> Is this going to do what I want? My hope would be in Lua that I could do `some_n9entity:name()` and get nil or its name if any.
> 
> Is this what we do now instead of implementing `UserData`?

Yup thats exactly right! Function context comes first, then any args, the method call syntax just desuggars to `fn(entity, arg1)` and the function context is sort of a virtual argument

makspll:
I'll have a crack at optionals next, rhai is now back :D

shanecelis:
What am I doing wrong with this? I can call `__map()` and return a `N9Entity` but when I call `retain()` I get the below error. 

```rust
#[derive(Clone, Reflect)]
pub struct N9Entity {
    pub entity: Entity,
    pub drop: DropPolicy,
}

pub(crate) fn register_script_functions(app: &mut App) {
    NamespaceBuilder::<N9Entity>::new(app.world_mut())
        .register("retain", |ctx: FunctionCallContext, this: Val<N9Entity>| {
            let world = ctx.world()?;
            world.with_global_access(|world| {
                let mut commands = world.commands();
                commands.entity(this.entity).remove::<Clearable>();
            })?;
            Ok(this)
        })
        ;
}
// ...
.register(
    "__map",
    |ctx: FunctionCallContext| {
        let id: Entity = // ...
            let entity = N9Entity { entity: id, drop: DropPolicy::Nothing };
            let world = ctx.world()?;
            let reference = {
                let allocator = world.allocator();
                let mut allocator = allocator.write();
                ReflectReference::new_allocated(entity, &mut allocator)
            };
            Ok(ReflectReference::into_script_ref(reference, world)?)
    },
)
```

```lua
    if my_map == nil then
        my_map = map()
        my_map:retain()
    end
```

```
ERROR bevy_mod_scripting_core::handler: error in script `../carts/nano9/cardboard-toad.p8#lua`: Could not find function: get for type: bevy_mod_scripting_core::bindings::reference::ReflectReference.
Context:
stack traceback:
        [C]: in metamethod 'index'
        [string "/Users/shane/Projects/bevy_mod_scripting/crat..."]:1693: in field 'draw'
        [string "/Users/shane/Projects/bevy_mod_scripting/crat..."]:1735: in method 'draw'
        [string "/Users/shane/Projects/bevy_mod_scripting/crat..."]:1837: in function '_draw'
Event handling for: Language: Lua
```

makspll:
That's strange, the error suggests the fallback for the index metamethod is failing, it normally looks for a "get" function in the registry to index into the type with, I.e. to get the properties.

Do you manipulate the script function registry? It may be possible you remove it at some point and forget to insert it, which causes the guard to create a default empty one

shanecelis:
I have an old `UserData` impl for it. I'll try taking that out.

shanecelis:
Confirm option arguments are working. Woohoo!

makspll:
> EDIT: I removed that. No change. My struct isn't a Component, Resource, or anything. That shouldn't matter, right?

Yeah it should not matter. It definitely looks to me like a symptom of a removed `AppScriptFunctionRegistry`

shanecelis:
I got a nice new error with a type id for the missing "get" or "index" function, so I went to confirm that it was my type.

```rust
pub(crate) fn plugin(app: &mut App) {
    info!("N9Entity type id {:?}", TypeId::of::<N9Entity>());
    NamespaceBuilder::<N9Entity>::new(app.world_mut()).register(
        "retain",
        |ctx: FunctionCallContext, this: Val<N9Entity>| {
            let world = ctx.world()?;
            world.with_global_access(|world| {
                let mut commands = world.commands();
                commands.entity(this.entity).remove::<Clearable>();
            })?;
            Ok(this)
        },
    );
}
```
Guess what didn't get logged? Yep, that `info!` line because it wasn't being called. [Slaps forehead.] Oy vey! Sorry for troubling you about it.

makspll:
It's all good, this is useful to me for improving the docs!

PR #223, opened 2025-01-22
feat: Add world.with_or_insert_component_mut()

shanecelis:
Just a convenience method I found useful.

I removed some panic comments on the with_component_*() because I don't believe they panic, but I didn't affirm that with tests.


makspll:
Looks good! The only thing I'd say is the function could be slightly more lenient in terms of claiming access when the component does exist (in which case it could just claim access to it) but I am not too bothered atm and that could be changed later

PR #250, opened 2025-02-08
feat: add recipient for specific language

pro470:
This is useful when you have BMS as a plugin system and want multiple languages implemented. You also want the events for different languages to be sent on different schedules. You need all language scripts to be on a dedicated entity and can not put them all on one “Language” entity.

### DIFF
--- a/crates/bevy_mod_scripting_core/src/event.rs
+++ b/crates/bevy_mod_scripting_core/src/event.rs
@@ -122,6 +122,8 @@ pub enum Recipients {
     Script(ScriptId),
     /// The event is to be handled by all the scripts on the specified entity
     Entity(Entity),
+    /// The event is to be handled by all the scripts of one language
+    Language(crate::asset::Language),
 }
 
 /// A callback event meant to trigger a callback in a subset/set of scripts in the world with the given arguments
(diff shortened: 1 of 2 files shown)


pro470:
This allows also us to have the luascriptplugin and all the system it needs in one plugin. And can create a more creative schedules with it

makspll:
Hi and thanks for contributing 👋!

This is a good point, I wonder if event handlers should by default only handle the language their plugin is parameterized by anyway? I.e. 

```rust
if P::Language != script.language {
    continue
}
```

Joakker:
> This is a good point, I wonder if event handlers should by default only handle the language their plugin is parameterized by anyway? 

That would be the logical option. However it would also necessitate keeping track of which ScriptId has which language

pro470:
I think this would not solve my issue because when I want all systems also the event send systems to be in one plugin I can't send dedicated events to a specific language 
This was also the problem I had. When I have event send systems in each language the first event handler system would run the Lua script twice because new_for_all sends it to all languages. With the recipient, I can have the send systems in their plugin but only for the language 
So the problem I have is not on the event handler but rather on the event being sent

makspll:
That makes sense, happy to add merge this then and if people have issues with this approach we can re-visit.

Could you run `cargo fmt` to pass CI?

pro470:
> That makes sense, happy to add merge this then and if people have issues with this approach we can re-visit.
> 
> Could you run `cargo fmt` to pass CI?

I'm happy to help when people have Issues. 😁
I'm also completely open to have a different implementation for this solution as well.

Issue #247, opened 2025-02-02
Design & Write language agnostic bindings format


makspll:
Given metadata and information about types functions and arguments, we should be able to produce "bindings" definition files, let's call them `language agnostic definition` files or LAD's.

These should contain everything we might need for generating:
- documentation 
- language server definition files
- lints

I am imagining something along the lines of:

```json
{
 "types": [
   {
     "name": "MyType",
     "id": 1234
     "path": "my_crate::mod",
     "generics": ["T"],
     "docstring": "A type that does type things...",
     "methods": [
       0,4,6
     ],
     "fields": [
        ...
     ],
     "variants": [
       ...
    ]
   }
 ],
 "functions": [
    "name": "my_func",
    "path": "my_crate::mod::func",
    "docstring": "a function that does function things...",
    "arguments": [
        {
            "type": 1234,
            "name": "my_arg",
            "kind": "Ref"
        }
    ]
 ]
}
```

makspll:
So far I have:
```json

{
  "version": "0.1.0",
  "types": {
    "bevy_mod_scripting_core::docgen::ladfile::test::EnumType": {
      "identifier": "EnumType",
      "crate": "bevy_mod_scripting_core",
      "path": "bevy_mod_scripting_core::docgen::ladfile::test::EnumType",
      "layout": [
        {
          "kind": "Unit",
          "name": "Unit"
        },
        {
          "kind": "Struct",
          "name": "Struct",
          "fields": [
            {
              "name": "field",
              "type": "usize"
            }
          ]
        },
        {
          "kind": "TupleStruct",
          "name": "TupleStruct",
          "fields": [
            {
              "type_": "usize"
            },
            {
              "type_": "alloc::string::String"
            }
          ]
        }
      ]
    },
    "bevy_mod_scripting_core::docgen::ladfile::test::StructType<usize>": {
      "identifier": "StructType",
      "crate": "bevy_mod_scripting_core",
      "path": "bevy_mod_scripting_core::docgen::ladfile::test::StructType<usize>",
      "generics": [
        {
          "type_id": "usize",
          "name": "T"
        }
      ],
      "documentation": " I am a struct",
      "layout": {
        "kind": "Struct",
        "name": "StructType",
        "fields": [
          {
            "name": "field",
            "type": "usize"
          },
          {
            "name": "field2",
            "type": "usize"
          }
        ]
      }
    },
[...]

PR #274, opened 2025-02-11
feat: create `Language Agnostic Declaration` file format and `ladfile` crate

makspll:
Closes: #247


makspll:
`ladfile` seems to have to be in its own crate, or otherwise the `serde_json` dependency in core causes issues when it's mixed in with bevy during binding generation

PR #294, opened 2025-02-16
feat: Add `TypedThrough` and `IntoScript` derive macros

Joakker:
Adds macros for the `TypedThrough` and `IntoScript` traits. Let me know if there are other traits with reasonable default implementations


makspll:
Looks good! Could you have a look at the clippy failures?

Joakker:
Will do

Joakker:
I think that should do it. The failing tests have to do with Ladfiles, which I didn't touch

makspll:
Yup, thats fine, thank you!

PR #302, opened 2025-02-22
feat: add `construct` global for constructing arbitrary types & `Union` type

makspll:
Makes this possible:
```lua
local type = world.get_type_by_name("SimpleEnum")

-- Struct Variant
local constructed = construct(type, {
    variant = "Struct",
    foo = 123
})

assert(constructed:variant_name() == "Struct", "Value was constructed incorrectly, expected constructed.variant to be Struct but got " .. constructed:variant_name())
assert(constructed.foo == 123, "Value was constructed incorrectly, expected constructed.foo to be 123 but got " .. constructed.foo)


-- TupleStruct Variant
local constructed = construct(type, {
    variant = "TupleStruct",
    _1 = 123
})

assert(constructed:variant_name() == "TupleStruct", "Value was constructed incorrectly, expected constructed.variant to be TupleStruct but got " .. constructed:variant_name())
assert(constructed._1 == 123, "Value was constructed incorrectly, expected constructed._1 to be 123 but got " .. constructed._1)

-- Unit Variant
local constructed = construct(type, {
    variant = "Unit"
})

assert(constructed:variant_name() == "Unit", "Value was constructed incorrectly, expected constructed.variant to be Unit but got " .. constructed:variant_name())
```

### DIFF
--- a/crates/bevy_mod_scripting_core/Cargo.toml
+++ b/crates/bevy_mod_scripting_core/Cargo.toml
@@ -48,6 +48,7 @@ bevy_mod_scripting_derive = { workspace = true }
 [dev-dependencies]
 test_utils = { workspace = true }
 tokio = { version = "1", features = ["rt", "macros"] }
+pretty_assertions = "1.4"
 
 [lints]
 workspace = true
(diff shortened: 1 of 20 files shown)


makspll:
There's a gotcha with tuples, which can be constructed but need `FromReflect` manually registered https://github.com/bevyengine/bevy/issues/17981

PR #328, opened 2025-02-26
feat: add test for construct using unit struct

mgi388:
Pushing this PR up so we can document how to construct unit structs. The first commit of this PR will fail with:

```
---- script_test - lua - /construct/construct_unit_struct.lua ----
error: Error in function construct : Error converting argument 2: Value mismatch, expected: std::collections::HashMap<alloc::string::String, bevy_mod_scripting_core::bindings::script_value::ScriptValue, std::collections::hash_map::RandomState>, got: [].
Context:
stack traceback:
        [C]: in function 'construct'
        [string "crates/languages/bevy_mod_scripting_lua/src/l..."]:2: in main chunk
```

Not sure how to get this to work without passing dummy values like this:

```lua
local type = world.get_type_by_name("UnitStruct")
local constructed = construct(type, {
    dummy = true
})

assert(constructed ~= nil, "Value was not constructed")
```

@makspll feel free to push to my PR if you have a fix. Hopefully this test documents this for future users.

### DIFF
--- a/crates/languages/bevy_mod_scripting_lua/tests/data/construct/construct_unit_struct.lua
+++ b/crates/languages/bevy_mod_scripting_lua/tests/data/construct/construct_unit_struct.lua
@@ -1,0 +1,4 @@
+local type = world.get_type_by_name("UnitStruct")
+local constructed = construct(type, {})
+
+assert(constructed ~= nil, "Value was not constructed")
--- a/crates/languages/bevy_mod_scripting_lua/tests/data/construct/construct_unit_struct.lua
+++ b/crates/languages/bevy_mod_scripting_lua/tests/data/construct/construct_unit_struct.lua
@@ -1,4 +1,4 @@
 local type = world.get_type_by_name("UnitStruct")
-local constructed = construct(type, map({}))
+local constructed = construct(type, {})
 
 assert(constructed ~= nil, "Value was not constructed")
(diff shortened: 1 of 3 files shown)


makspll:
Ah you've hit an interesting case!
At the moment `ScriptValue` has both a Vec and a Map variant,
Now during conversion from lua, mlua has to know the target type to convert to, the way it currently does that, is by looking at the type of keys in your table.

Obviously lua doesn't actually have a distinction between maps and lists, which is the root cause here!

One way to deal with this would be to introduce a `map({..})` constructor, which explicitly takes in either a `Vec` or a `HashMap`. 
Another way would be to introduce a new `FromScript` implementing primitive `ListOrMap`, I guess you can already do this with `Union`, and use that instead of Map or List everywhere, but I think the map way is less restrictive on binding authors

makspll:
Once #329 merges this should work

makspll:
Yes that would be appreciated

makspll:
Oh never mind,  because it's converting to a script value it's getting back to a table in lua, I need to do something else

makspll:
Okay, alternatively, we can make hash maps FromScript impl allow lists of tuples as well, that way it will be happy with an empty list

makspll:
https://github.com/makspll/bevy_mod_scripting/pull/332 will definitely do the trick and without `map`

PR #341, opened 2025-03-03
Set get hashmap

hakolao:
Isn't `insert` basically same as `set` for maps?

Would something like this work?


makspll:
Closing as this should now be somewhat easier to achieve with hte new reference system

PR #345, opened 2025-03-03
fix!: script contexts being completely overwritten on a re-load

makspll:
# Summary
- Currently on a script re-load, the entire context will be replaced for both lua and rhai
- This does not work well with shared context mode
- There probably isn't any harm to not doing this normally either
- I.e. scripts shouldn't hold much global state, and if they do it might be beneficial to re-load while retaining that state
- Ultimately the script lifecycle should take care of re-freshing the state correctly
- This PR makes script re-loads merge the new and old contexts instead of wiping the old one and installing the new one


makspll:
Looks like this broke rhai

PR #357, opened 2025-03-07
fix: Sandboxing Scripts in Shared Contexts

Peepo-Juice:
# Summary
This PR changes the way scripts are loaded and how event handlers are called so that scripts don't collide when sharing the same context. Works to solve #339.

### Load
- **Before:** We would simply load the script without setting an environment, thus populating the global scope of the language context.
- **Now:** An environment is created and set when loading the script, thus populating this new environment instead of the global scope. We are also injecting the context's globals into this environment, thus maintaining the script's access to the globals it expects to have. We store this new environment in a `ScriptEnvironmentStore`.

### Function Calls
- **Before:** We would search the context's globals for the requested function and then call it.
- **Now:** We grab the environment from the `ScriptEnvironmentStore` via ScriptId, search for the requested function in that environment and then call it.

# Next Steps
- We need to apply this for Rhai and Rune, so far I only did Lua
- Discuss whether we should apply this sand-boxing all the time, or only when shared context is enabled? Currently its doing it all the time, which didn't break anything for me personally when I turned off context sharing, but might have other implications I'm not aware of.
- Write tests to ensure that two scripts that have the same event handlers can be triggered at will with no collisions, and still have access to data stored as globals (including BMS type registrations and other bindings)
- Cleanup, I don't think I did a good job with the documentation.


Peepo-Juice:
Regarding the "when to sandbox": before my changes, the globals of scripts were accessible by every other script when context sharing was disabled. So this causes a "breaking change" in that respect. However, you mentioned in Discord that your stance was that scripts should be private by default, but you may have meant strictly for shared contexts.

crates/bevy_mod_scripting_core/src/asset.rs
```diff
@@ -18,6 +18,7 @@ use bevy::{
     reflect::TypePath,
     utils::HashMap,
 };
+use mlua::Table;
```
makspll, line 21: 🙅 This is introducing lua logic into the bms_core layer. The idea is to only keep language agnostic logic in here

Peepo-Juice:
I will be fixing the issue of event handler collisions via a different method. I'll let someone else work on this later :)

PR #358, opened 2025-03-08
feat: Allow ScriptComponent to be inserted from lua scripts

Peepo-Juice:
Just needed to call register_type on ScriptComponent and add `#[reflect(Component)]`.


Peepo-Juice:
Sorry, I messed up. It included changes that I didn't intend to include.

Issue #419, opened 2025-05-21
Allow state to be passed between reloads


shanecelis:
# Problem

When I'm editing a script and the game is running, it resets all the state upon reload.

# Solution

I'd like there to be some mechanism to pass state before a script is unloaded to the reloaded script.

I initially thought overloading the `on_script_loaded()` could accept a state argument and `on_script_unloaded()` could provide state. 

# Alternative Solution

But discussion in the discord suggested maybe a particular variable ought to be designated be preserved between scripts.

I was thinking about trying my hand at implementing this, but I haven't done anything yet.

PR #421, opened 2025-05-21
feat: Add on_script_reloaded callback, allow retaining state between reloads

shanecelis:
# Summary

I added an `on_script_reloaded` function that is called before and after on_script_loaded on reloads. This addresses the issue I filed #419.

I updated the core-callbacks doc with example code. Here's the example code shown there:

```lua
mode = 1
function on_script_reloaded(save, value)
    if save then
        print("Before I go, take this.")
        return mode
    else
        print("I'm back. Where was I?")
        mode = value
    end
end
```

I'm not certain about the name. `on_script_reload` might be the name I'd choose with a blank slate. 

## Alternatives

One could designate a special variable to capture. But I think a function is preferable because a reload can interrupt at any time and keeping that value up to date all the time instead of when it's required seems inefficient.

One could break this into two functions, but they don't do anything unless both are implemented so it seemed wise to keep them as a single function.

## Testing

I exercised it manually with errors in both the saving of state and restoring state, and both were reported as the usual mechanism.

## Mistake

I started this branch on top of my other pull request #418. Oops. Luckily #418 is not controversial or large.

### DIFF
--- a/crates/bevy_mod_scripting_core/src/commands.rs
+++ b/crates/bevy_mod_scripting_core/src/commands.rs
@@ -6,7 +6,7 @@ use crate::{
     context::ContextBuilder,
     error::{InteropError, ScriptError},
     event::{
-        CallbackLabel, IntoCallbackLabel, OnScriptLoaded, OnScriptUnloaded, OnScriptReloaded,
+        CallbackLabel, IntoCallbackLabel, OnScriptLoaded, OnScriptReloaded, OnScriptUnloaded,
         ScriptCallbackResponseEvent,
     },
     extractors::{with_handler_system_state, HandlerContext},
@@ -273,10 +273,10 @@ impl<P: IntoScriptPluginParams> Command for CreateOrUpdateScript<P> {
                     OnScriptReloaded::into_callback_label(),
                     vec![ScriptValue::Bool(false), state],
                     false,
-                ).apply(world);
+                )
+                .apply(world);
             }
         }
-
     }
 }
 
--- a/crates/bevy_mod_scripting_core/src/commands.rs
+++ b/crates/bevy_mod_scripting_core/src/commands.rs
@@ -212,7 +212,7 @@ impl<P: IntoScriptPluginParams> Command for CreateOrUpdateScript<P> {
                                         vec![err
                                             .with_script(self.id.clone())
                                             .with_context(P::LANGUAGE)
-                                            .with_context("saving reload state")]
+                                            .with_context("saving reload state (shared-context)")]
                                         .into_iter(),
                                     );
                                 }
@@ -222,6 +222,27 @@ impl<P: IntoScriptPluginParams> Command for CreateOrUpdateScript<P> {
                         self.reload_context(guard.clone(), handler_ctxt)
                     }
                     None => {
+                        match handler_ctxt.call_dynamic_label(
+                            &OnScriptReloaded::into_callback_label(),
+                            &self.id,
+                            Entity::from_raw(0),
+                            vec![ScriptValue::Bool(true)],
+                            guard.clone(),
+                        ) {
+                            Ok(state) => {
+                                reload_state = Some(state);
+                            }
+                            Err(err) => {
+                                handle_script_errors(
+                                    guard.clone(),
+                                    vec![err
+                                        .with_script(self.id.clone())
+                                        .with_context(P::LANGUAGE)
+                                        .with_context("saving reload state")]
+                                    .into_iter(),
+                                );
+                            }
+                        }
                         bevy::log::debug!("{}: loading script with id: {}", P::LANGUAGE, self.id);
                         self.load_context(guard.clone(), handler_ctxt)
                     }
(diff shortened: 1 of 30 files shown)


shanecelis:
## A further alternative 

I guess as an alternative going with the two functions, one could do this:

```lua
mode = 1
function on_script_reload()
    print("Before I go, take this.")
    return mode
end

function on_script_loaded(value)
    if value then
        print("I'm back. Where was I?")
        mode = value
    end
end
```
My criticism above doesn't apply because `on_script_loaded()` is useful without `on_script_reload()`. We'd then only have this set of function calls when reloading:

1. on_script_loaded()
2. INITIATE RELOAD
3. value = on_script_reload()
4. CONTEXT CLEARED
5. on_script_loaded(value)

The preceding design looks like this for comparison:

1. on_script_loaded()
2. INITIATE RELOAD
3. value = on_script_reload(true)
4. CONTEXT CLEARED
5. on_script_loaded()
6. on_script_reload(false, value)

shanecelis:
NOTE: This currently only works when context sharing is enabled. I made an attempt to support non-shared contexts, but it seemed potentially infeasible. I'd love to hear @makspll's take on whether that seems right.

crates/bevy_mod_scripting_core/src/commands.rs
```diff
@@ -194,6 +195,28 @@ impl<P: IntoScriptPluginParams> Command for CreateOrUpdateScript<P> {
                             // it can potentially be loaded but without a successful script reload but that
                             // leaves us in an okay state
                             handler_ctxt.scripts.scripts.insert(self.id.clone(), script);
+                        } else {
+                            match handler_ctxt.call_dynamic_label(
+                                &OnScriptReloaded::into_callback_label(),
+                                &self.id,
+                                Entity::from_raw(0),
+                                vec![ScriptValue::Bool(true)],
```
makspll, line 203: ⚠️ for languages like rhai, where the arity of the callback is used to actually identify the function in the script, we would probably want to pass in the same number of arguments both times, but instead pass in a `None` for the missing argument

docs/src/ScriptingReference/core-callbacks.md
```diff
@@ -30,3 +31,24 @@ function on_script_unloaded()
...
 end
 ```
+
+## `on_script_reloaded`
+
+This will be called twice: right before and after a script is reloaded.
+
+The first parameter `save` informs you whether it is time to save a value or restore it.
+
+Before the reload, it is called with one argument: `true`. After the script is reloaded, it is called with two parameters: the first is `false` and the second is value returned from before.
```
makspll, line 41: as of now this seems to be misleading for non-shared contexts, in which case only the second half of the callback would happen

shanecelis:
> * it seems to me this should be possible to make work in non-shared contexts too
>   
>   * what were the problems you faced here?

Let me see if I can recreate the problem I faced. Huh. Well, that's funny. It's working now and it's as easy as I hoped it would be. New PR #424 incoming. 

The block of code that calls the `on_script_reloaded(true)` is virtually identical for both cases and maybe can be extracted into one block. The shared-context case had some conditional code that suggested it made sense to only call it in one of its cases. It works now for both cases, but I'll leave it to your more practiced hands on this code base if the code duplication ought to be refactored.

>   * having these callbacks work slightly differently between the two modes of bms would be unnecessarily confusing IMO

I agree. It'd be best to unify them.

docs/src/ScriptingReference/core-callbacks.md
makspll: ohh, I didn't catch the second conditional

crates/bevy_mod_scripting_core/src/commands.rs
shanecelis: Oh, I did not know that. Yes, let me make it arity invariant then.

shanecelis:
I am sorry. I got confused because you merged main into _my branch_, which is fine, but I interpreted it as this PR had gotten merged into _your repo_, which isn't the case. That's why for a follow up I created the #424 PR. Ugh. Sorry about that. Let me try to get back on track with this PR.

crates/bevy_mod_scripting_core/src/event.rs
```diff
@@ -73,8 +74,13 @@ macro_rules! callback_labels {
 }
 
 callback_labels!(
+    /// Fired when a script is successfully loaded
     OnScriptLoaded => "on_script_loaded",
+    /// Fired when a script is unloaded before a reload, if a value is returned, it will be passed to the `on_script_reloaded` callback
     OnScriptUnloaded => "on_script_unloaded",
+    /// Fired when a script is reloaded (loaded after being unloaded)
+    /// This callback receives the value returned by the `on_script_unloaded` callback if any were returned
+    OnScriptReloaded => "on_script_reloaded",
```
makspll, line 83: Split into unloaded and reloaded

crates/bevy_mod_scripting_core/src/commands.rs
```diff
@@ -145,104 +148,174 @@ impl<P: IntoScriptPluginParams> CreateOrUpdateScript<P> {
...
+                vec![state],
+                false,
+            )
+            .with_context(P::LANGUAGE)
+            .with_context("on reloaded callback")
+            .run_with_handler(world, handler_ctxt);
+        }
+    }
+
+    fn handle_global_context(
```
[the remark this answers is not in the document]
shanecelis: Oh, this is much nicer. Good call.

PR #422, opened 2025-05-22
chore!:Bevy 0.16

alexparlett:
# Summary
This introduces Bevy 0.16 compatibility.

We should probably in a follow up move to the bevy::platform::collections instead of using std to try and align as closely as possible with the no std approach bevy is taking but i didn't want to do it in this one.

Need to quickly restore the ordering where my ide aligned them with the interface order, but thought it would be good to get visibility on the changes.


CrushedPixel:
Is there anything holding this up? Formatting needs to be fixed up, anything else to do?

Nub:
Bump on getting this landed?

makspll:
Looking at this now, it looks like there is still some work left on the codegen side which is currently failing. Did you update the generated bindings manually 👀

shanecelis:
I didn't. And don't know how. Will look for your commit as instructive for next time.

makspll:
The good news is it seems the new rustup version didn't break the codegen crate (i.e. it does not require a `rustc_plugin` version bump, which is always the most painful bit)

Codegen seems to need a few small adjustments and we should be good to go

makspll:
I've had my yearly fight with rust analyzer now, a few little changes to go

makspll:
FYI to those who have been using this branch, bevy shuffled around many types, which will be reflected in the new codegen, things that used to live in `bevy_core`, and `bevy_hierarchy` like `Name` seem to have been moved to `bevy_ecs`

crates/bevy_mod_scripting_core/src/reflection_extensions.rs
```diff
@@ -424,11 +425,19 @@ impl<T: PartialReflect + ?Sized> PartialReflectExt for T {
...
         reflect: &dyn PartialReflect,
         world: WorldGuard,
-    ) -> Box<dyn PartialReflect> {
+    ) -> Result<Box<dyn PartialReflect>, InteropError> {
         // try from reflect
         match <dyn PartialReflect>::from_reflect(reflect, world.clone()) {
-            Ok(v) => v.into_partial_reflect(),
-            Err(_) => reflect.clone_value(),
+            Ok(v) => Ok(v.into_partial_reflect()),
+            Err(_) => reflect
```
makspll, line 432: I believe this is what we ultimately wanted this function to do, I can't remember if allowing non-fully-reflect values to be returned here was a convenience thing and it just never happened or if it was used for something at specific at some point, either way, if it causes trouble we can have a look

makspll:
See: https://github.com/bevyengine/bevy/issues/20395#issuecomment-3180870538 for the above, TL;DR the bevy update causes unrelated commands to flush when world entities or components are modified

PR #424, opened 2025-05-26
Add reload hook support for non-shared contexts.

shanecelis:
This is an addendum to PR #421. Let's keep the discussion there.


shanecelis:
Closed. I only opened this because I thought #421 had been merged. Its contents will be merged there.

Issue #426, opened 2025-06-27
ScriptAsset loading is script evaluation, and why Script IDs?


shanecelis:
Firstly, I love bevy\_mod\_scripting. It's a huge part of why I thought the prospect of [Nano-9](https://github.com/shanecelis/nano-9) seemed feasible. As an avid user I've developed some opinions. Let me relate them as issues.

# Problem: ScriptAsset loading is script evaluation

I've got two problem instances for this general problem.

## Problem Instance 1: Multiple scripts

In Nano-9 I'm trying to support someone providing multiple scripts in a shared context environment like this:
```toml
scripts = [
"vector.p8", // p8 is just the Pico-8 dialect of Lua, which I translate into vanilla Lua.
"actor.p8",
"platformer.p8",
"dolly.p8",
"main.lua",
]
```
This happens through an `AssetLoader`, which is restricted environment; it's an async function where I don't have access to the world or resources, etc. Some of the scripts do depend on the preceding scripts. Although I load these `ScriptAsset`s sequentially using an immediate loader, they are evaluated in an indeterminate order, which often means the wrong order. 

## Problem Instance 2: Can't setup the environment fast enough

I'd really like to be able to support having a hello world script that looked like this:
```lua
print("hello world")
```
However, because of this issue, it must currently look like this:
```lua
function _init()
   print("hello world")
end
```
The reason is that in my config file where I load the code, I also load the sprite sheets and color palette. I need to set up the color palette before I can call `print()`.

## Solution A: Don't do that.
I could punt the loading of scripts outside of the `AssetLoader` but images, sounds, fonts, etc. load this way, why not scripts?

## Solution B: Separate script evaluation and script loading.
Let's evaluate scripts at another time, perhaps when they're added to a ScriptComponent.

## Solution C: Evaluate scripts in order of load.
If we can't change the evaluation after loading can we ensure the order they're evaluated is the order they were loaded?

# Problem: ScriptIds complicate everything
The script IDs have been a hassle for me to get right. I’ve added and removed and changed my asset path to script id mapper many times. And when it stops working, it’s hard for me to figure out why. 

> As an aside, I added Reflect to everything once so I could see what was going on in the inspector. (This was before it was part of BMS proper. Love seeing a changes that mimic my own homebrew!)

When I take a step back I mostly don’t understand why script IDs exist. 

[...]

PR #427, opened 2025-06-27
feat: Use the Handles, Luke!

shanecelis:
I was only planning to file issue #426 tonight, which states the problems I was having. I thought if I poked around I'd find the implementation reason for why what I suggest there wouldn't work. So I made a branch and rolled up my sleeves, and I was able to pretty quickly get exactly what I wanted.

# Handles
In the issue I suggest doing away with `ScriptId`. As I worked with the implementation, it became clear I actually wanted to redefine it.
```rust
// OLD
// type ScriptId = Cow<'static, str>; 
// NEw
type ScriptId = AssetId<ScriptAsset>;
```
Beyond that I made it so `ScriptComponent` holds `Handle<ScriptAsset>`s. The nice thing about this is you can use this pattern:
```rust
ScriptComponent(vec![asset_server.load("foo.lua")])
```
No need to store a strong handle somewhere so your script isn't unloaded. However, if you rely on that behavior you can still do that like you do with any other asset-based item using weak handles.

```rust
let strong_handle = asset_server.load("foo.lua");
ScriptComponent(vec![strong_handle.clone_weak()])
```
This preserves the semantics of the current release.

# ScriptAsset loading versus evaluation
In this PR the static scripts are ~~handled essentially the same as before except they~~ can retain handles (both strong and weak). EDIT: A script may be loaded without evaluation. If it is added as a static script, it will then be evaluated.

The non-static scripts are not evaluated until they are added to a `ScriptComponent`. Then they are evaluated in order, sequentially.

# ScriptAsset change
I suggested a change to `ScriptAsset` and that survived this refactoring.

```rust
pub struct ScriptAsset {
    /// The body of the script
    pub content: Box<[u8]>,
    /// The language of the script
    pub language: Language,
}
```
I also provided a `ScriptSettings` where one can provide a definite language. `None` will use the file's extension. 
```rust
asset_server.load_with_settings("hello.p8", |settings: &mut ScriptSettings| {
   settings.language = Some(Language::Lua);
});
```
One downside about `ScriptSettings` is it required adding "serde" with the "deriv" feature as a dependency.

# Example

I updated the Game of Life to work with Lua and Rhai and all its console options.

# Remaining work

I have not tried to exercise any other languages or even do much with the tests. But I'd be happy to move this forward to handle the rest with some consensus that that's the right direction.


shanecelis:
I got the tests compiling and running. There are 9 failures with features 'lua' and 'rhai'.

shanecelis:
There is a regression in that it doesn't support multiple scripts. I'm trying to get that ready in the 'use-entities' branch, which is compiling now but not running yet if you wanted to take a peek. But maybe this PR needs to cook a little more. I'm gonna take another pass on it tonight.

shanecelis:
Ok, I have fixed the regression. Game of life works in all ways for Lua. For Rhai it doesn't seem to work for static scripts.

So as I was trying to make scripts work with a shared context and per entity/script, I realized the split of contexts is actually a big policy decision. I imagine in most cases it'll come down to a few qualities: entity, script, and another thing I'm calling domain. To that I end I wrote this trait:

```rust
/// A generic script context provider
pub trait ScriptContextProvider<P: IntoScriptPluginParams> {
    /// Get the context.
    fn get(&self, id: Option<Entity>, script_id: &ScriptId, domain: &Domain) -> Option<&Arc<Mutex<P::C>>>;
    /// Insert a context.
    fn insert(&mut self, id: Option<Entity>, script_id: &ScriptId, domain: &Domain, context: P::C) -> Result<(), P::C>;
    /// Returns true if there is a context.
    fn contains(&self, id: Option<Entity>, script_id: &ScriptId, domain: &Domain) -> bool;
}
```
## SharedContext 
So the shared context is implemented with this:
```rust
pub struct SharedContext<P: IntoScriptPluginParams>(pub Option<Arc<Mutex<P::C>>>);
```
and it ignores the arguments and just provides the same context to everyone.
## EntityContext
A per entity context is implemented like this:
```rust
/// Stores the script context by entity.
pub struct EntityContext<P: IntoScriptPluginParams>(HashMap<Entity, Arc<Mutex<P::C>>>);
```
## ScriptContext composition
And the resource we make available is a composition:
```rust
#[derive(Resource)]
/// Keeps track of contexts
pub enum ScriptContext<P: IntoScriptPluginParams> {
    /// One shared script context
    Shared(SharedContext<P>),
    /// One script context per entity
    ///
    /// Stores context by entity with a shared context as a last resort when no
    /// entity is provided.
    Entity(EntityContext<P>, SharedContext<P>)
}
```
## Domain
But what's domain? It's the old `ScriptId` sort of, maybe?

```rust
/// A kind of catch all type for script context selection
///
/// I believe this is what the original ScriptId was intended to be.
pub type Domain = Option<Cow<'static, str>>;
```
My hope is this would allow us to provide the right criteria to choose contexts divisions. I don't do anything with `domain` presently, and if you think it's extraneous we can toss it. 

My vision for domain was to allow users to clump scripts together a little more haphazardly. Something like the following fantasy code would put scripts into one of two domains:
```rust
asset_server.load_with_settings("player.lua", |settings: &mut ScriptSettings| {
   settings.domain = Some("player".into());
});
asset_server.load_with_settings("monster.lua", |settings: &mut ScriptSettings| {
   settings.domain = Some("env".into());
});
asset_server.load_with_settings("effects.lua", |settings: &mut ScriptSettings| {
   settings.domain = Some("player".into());
});
```
## Take away
This code is working and I think it's ready for your attention.

Having spent more time with this code base I'm in even more awe of what you've put together. Let me know if there's anything I can do to help. I'm sorry that this branch is probably going to conflict with 0.16 branch. :(

## Further work

I'm going to try to make this branch work with Nano-9 to dog food it in the coming days.

shanecelis:
## Domains

I went ahead and fleshed out the `Domain` idea and dropped the `Option` from its
type, but in almost every parameter it is an `Option<Domain>` now.

```rust
pub type Domain = Cow<'static, str>;
```
## Add support for context per script
I added `with_domains()` and `per_script()`. 

```rust
impl<P: IntoScriptPluginParams> ScriptContext<P> {
    /// Use a shared script context
    pub fn shared() -> Self {
        Self::Shared(SharedContext::default())
    }
    /// Domain contexts or a shared context
    pub fn with_domains() -> Self {
        Self::Domain(DomainContext::default(), SharedContext::default())
    }
    /// Use one script ontext per entity
    pub fn per_entity() -> Self {
        Self::Entity(EntityContext::default(), SharedContext::default())
    }
    /// Use one script ontext per entity
    pub fn per_script() -> Self {
        Self::ScriptId(ScriptIdContext::default())
    }
}
```

Now that there are 4 choices, I'd suggest we change the `enable_shared_context`
to instead pick a default. If the user wants to change it they do:

```rust
app.insert_source(ScriptContext::shared());
```

## Problem: N scripts produce N callback evaluations even with 1 shared context

I added a `Recipients::Domain` and patching everything up I dealt with an issue
I came across as soon as I started using multiple scripts in BMS. If I have five
scripts in a shared context, and I do a callback for `_update()` with
`Recipients::All`, `_update()` is called FIVE TIMES! One can argue that's
correct behavior, but I found it surprising. It's not surprising when every
script is in its own context. I worked around this issue in Nano-9 by just
taking note of one `ScriptId` and only calling that. However, updating the code
I found an opportunity to implement a general solution.

## Proposal: Evaluate Callbacks on a Per Context Basis

Here is my take for least-surprise: If I have five scripts all in one shared
context, and I fire a call to `Recipients::All`, then I want that call to go to
the shared context once.

If I have four scripts in three different contexts, and I fire a call to
`Recipients::All`, then I want that call to be evaluated three times,
once in each context.

To facilitate that, I added the following method:

```rust
/// A generic script context provider
pub trait ScriptContextProvider<P: IntoScriptPluginParams> {
    // ... 
    /// Hash for context.
    ///
    /// Useful for tracking what context will be returned by `get()` without
    /// requiring that `P::C` impl `Hash` and cheaper too.
    fn hash(&self, id: Option<Entity>, script_id: &ScriptId, domain: &Option<Domain>) -> Option<u64>;
}
```

What I think is nice about this is it works the same as before if you're placing
every script in its own context. But it will also work with scripts, domains,
and shared contexts.

## Docs?

Would you like some help updating the docs/book for this PR?

shanecelis:
I'm prepping this branch for a BMS 0.14.0. 

Expecting some real-world use I updated the `ScriptContext<P>` to allow for the following kinds:
- `ScriptContext::shared()`
- `ScriptContext::per_script()`
- `ScriptContext::per_entity()`
- `ScriptContext::domains()`
- `ScriptContext::shared().with_domains()`
- `ScriptContext::per_script().with_domains()`
- `ScriptContext::per_entity().with_domains()`

And its default is `ScriptContext::per_script()` so that it follows the conventions of its preceding versions.

I found some tests failing (and some not even compiling--eek). I'll try to fix them tomorrow.

shanecelis:
It's all compiling now. 3 tests are failing. Unfortunately the fix seems a little hairy. The called_on_right_recipients test in particular requires a change in the event handler to fix.

My workday has been interrupted quite a bit, so I put my partial commits into my "use-handles-dev" branch so it's not distracting in this PR. I changed the event handler and now 5 tests are failing. On it goes.

shanecelis:
All tests pass! The game of life example starts, stops with Lua & Rhai with and without static argument.

# Assumption Breakage

I had been adding `Option<Entity>` and `Option<Domain>` parameters and fields as necessary, but the assumption of one context per script handle was still embedded and becoming awkward. I decided to bite the bullet and break with that assumption with this struct:

```rust
pub struct ContextKey {
    /// Entity if there is one.
    pub entity: Option<Entity>,
    /// Script ID if there is one.
    pub script_id: Option<Handle<ScriptAsset>>,
    /// Domain if there is one.
    pub domain: Option<Domain>,
}
```
It's the new key to context.

```diff
/// A generic script context provider
pub trait ScriptContextProvider<P: IntoScriptPluginParams> {
    /// Get the context.
+    fn get(&self, context_key: &ContextKey) -> Option<&Arc<Mutex<P::C>>>;
-    fn get(&self, id: Option<Entity>, script_id: &ScriptId, domain: &Option<Domain>) -> Option<&Arc<Mutex<P::C>>>;
    // ...
}
```
# What's in a Domain?
With the inclusion of `Domain` in the `ContextKey`, I wanted it to be `Copy` if possible, so I altered its defintion to this:
```rust
/// A kind of catch all type for script context selection
#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
pub struct Domain(u64);

impl Domain {
    /// Create a domain handle.
    pub fn new(hashable: impl Hash) -> Self {
        Domain(DefaultHashBuilder::default().hash_one(hashable))
    }
}
```
# New is Old
This is the new default `ScriptContext` variant that captures the prior version's behavior and then some.
```rust
    DomainEntityScriptId(Or<DomainContext<P>, Or<EntityScriptIdContext<P>, Or<ScriptIdContext<P>, SharedContext<P>>>>),
```
You can read the `Or`s as short-circuiting like this:
1. If there is a domain, `DomainContext` provides a context, or
2. if there is an entity and script, `EntityScriptIdContext` is used, or
3. if there is a script (which is the case for StaticScripts that have no entity), `ScriptIdContext` is used, or
4. if there is no domain, entity, or script, then `SharedContext` is used.

The prior version had behaviors 2 and 3 and 4 as a separate assignment mode. This adds domains without interfering with the rest if they're not used.

And its easy to design different behaviors within the library. For instance, if we wanted to have all static scripts live in the same context we can drop the `ScriptIdContext`.

```rust
    DomainX(Or<DomainContext<P>, Or<EntityScriptIdContext<P>, SharedContext<P>>>),
```
This means `SharedContext` will provide contexts for any context key that only has a script ID like static scripts. (We could also make static scripts work with domains as an alternative that's more flexible.)

# Reify me not
I did a big refactor on the `event_handler_inner<P: IntoScriptPluginParams>` function to make it work with the ScriptContext. I chose some code duplication for hopefully better runtime performance. Is there any performance tests I could exercise to compare with the previous version?

It now has these benefits:

- Early exits
- Does not query and reify all scripts to a collection every call
- Queries one entity only when necessary
- Queries scripts only when necessary

# Apologies

I am sorry that this PR has gotten so big. At the beginning it seemed like this could be a less invasive set of changes. Having failed to make this change set small, I have endeavored to make it good.

shanecelis:
Great to hear!

I think the versions sound fine. A handles+bevy0.15 as an alpha is good by me, better even than polluting the version space with use-handles and not.

> rebase this branch against that [no-handles+bevy0.16], to get the handles+bevy0.16 version

I think the desired end is good, but I worry the process of rebasing this branch onto a 0.16 will be a chore. But maybe it won't. To update this branch to "main," I'm going to squash everything in this branch just to make the rebasing easier. Maybe it won't be as troublesome if the commits are fewer, but I was initially thinking I'd take the patch from the old handles+0.16 and apply that on top of this branch with whatever fix up was required.

shanecelis:
I squashed this branch and rebased it to main. It's compiling. There are two tests which are failing.
```
failures:
    commands::test::test_commands_with_default_assigner
    commands::test::test_commands_with_global_assigner
```
And I'm not seeing the unload and reload work yet. Probably need to add a test for those.

shanecelis:
Reload functionality is now working!

## Asset Handling Behavior Change

I forgot to mention this. But we've been using `ScriptAsset` handles as a proxy for their contexts. When we dropped all script handles, we could drop the context. Because of breaking the 1-to-1 relationship that doesn't seem workable anymore. Instead to specifically purge a context, we now treat both scripts and static scripts the same, in that, we call `DeleteScript`, which will delete the static script or delete a script's context. (Perhaps there should be a `DeleteContext` command?) I would suggest making a new system called `auto_delete_script` that will look for the script handle removals and call `DeleteScript` automatically to offer a transition path.

## TODO
- [ ] Write system to automatically delete entity contexts when associated entity despawns.
- [ ] Write auto_delete_script system.
- [ ] Write migration-guide.md.

makspll:
🚧 WIP: Review in progress🚧 

Alright, the PR changes quite a few things I am going to use the following convention in my reviews:
- 🤷 Things of note, not requesting any changes / starting a discussion that might lead to changes requested, would like to hear your thoughts
- 🛑 Things that will block merging, I would like to see changes before going further.

I'll also divide my thoughts into the areas that were touched:

<details>
<summary>🤷 Script asset events directly tied to script asset handler systems</summary>

https://github.com/shanecelis/bevy_mod_scripting/blob/67c5a5d08c5f88f1865849013ca292fd5d009b40/crates/bevy_mod_scripting_core/src/asset.rs#L61-L65

This was previously decoupled via proxy script events for two reasons:
 1. changes to bevy asset events were easier to migrate, and the logic downstream does not have to worry about the nuances within bevy's own events
 2. as a staging ground for the script metadata resources (i.e. we would link up all the script metadata at the points we would synchronize these events)

I believe this was removed in this PR since the metadata synchronising is no longer necessary, but for the sake of 1. I'd personally still see value in the decoupling, maybe via:

```rust
pub(crate) enum ScriptAssetEventKind {
    Added,
    Removed,
    Modified,
}
```
and a matching struct with the payload, or repeated enum constructors whatever's easier.

</details>

shanecelis:
Thanks for taking the time to look over it.

## Script asset events

> I believe this was removed in this PR since the metadata synchronising is no longer necessary

That was the reason, but I think you're right that `ScriptEvent`s should be reconstituted especially since now the events are not aligned with the `AssetEvent`.

## Script evaluation queue feels unnecessary 

It started as a system `Local` before I was compelled to pull it out for static scripts.

I realized too that the queue isn't necessary unless you're putting scripts into the same context.

I will attempt to do something as you suggest but I admit I'm foggy on what that will entail.

## ScriptContext enum is very complex

That's a really good point. It's perhaps a labor of understanding that we can now schluff off, but I think you're right. We could get the same effect by having a `HashMap<ContextKey, P::C>` and applying some filter to the ContextKey before hand:
```rust
fn shared(context_key: ContextKey) -> ContextKey {
    ContextKey::default()
}

fn per_script(context_key: ContextKey) -> ContextKey {
    ContextKey {
      script: context_key.script,
      ..default()
    }
}
```
Oh, I see perhaps you'd suggest `Fn(&ContextKey) -> ContextId` rather. It does seem like `ContextId` wants to be expressed somehow.

I kind of liked how script handles were a proxy for context handles—when all script handles were dropped, the context was dropped—and it has made me wonder if Contexts ought to be assets with their own handles. [Shrug.]

> The Domain concept you've come up with then appears for free, as you can map to arbitrary ID's based on the set of keys we make available 

Hmm, I don't quite follow but I'm intrigued by the idea. What's the type of `ContextId` in your mind?

> (assuming you'd be driving domain off the asset paths etc)

I hadn't thought domains would be related to asset paths. Best example I have for their potential use is what I used in the [book example](https://github.com/shanecelis/bevy_mod_scripting/blob/use-handles/docs/src/Summary/contexts.md):
```rust
commands.spawn((
    ScriptComponent(vec![asset_server.load("player.lua")]),
    ScriptDomain(Domain::new("player")),
)).with_children(|parent| {
    parent.spawn((
        ScriptComponent(vec![asset_server.load("sword.lua")]),
        ScriptDomain(Domain::new("player")),
    ));
})
```

shanecelis:
I completely reworked the ScriptContext, and I think it's much better now. I dropped all those ScriptContextProvider modules. I dropped the ScriptContextProvider trait. Took out a lot of lines of code. Yay!

We use a simple enum `ContextRule` that allows for custom rules. The user can now configure policies much more flexibly, e.g., `[ContextRule::Domain, ContextRule::Custom(custom_context_selection_fn), ContextRule::Script, ContextRule::Shared]`.

I've reintroduced the `ScriptEvent`, and `ScriptQueue` has been demoted to a `Local` and type alias (and this incidentally fixed a latent bug that would have become apparent when multiple runtimes were mucking with one ScriptQueue).

I haven't sorted out the tests yet. 

There is a complication with using the `ContextKey` directly as the key in `ScriptContext`: Because the `script` field is an `Option<Handle<ScriptAsset>>`, I worry that a strong handle vs a weak handle for the same asset may not hash or == as we might like. I need to either ensure we only store weak handles or validate that the operations for search and insert won't depend on the handle being strong or weak. In my 'use-handles-dev' I have a commit that changes `ContextKey.script` to be a `Option<ScriptId>` but then we'll lose a lot of the nicer error reporting that can show you the path of the script during logging.

shanecelis:
All right. This is hopefully looking more like what you asked.

`DeleteScript` is kind of doing double duty. It deletes a static script and it deletes a context. Perhaps a `DeleteContext` should be extracted out from it. I don't know.

Also since we're storing `ContextKey`s as our key, I'll just note I'm only storing weak handles for scripts.

makspll:
Looking really good! The events make this extremely easy to read!

makspll:
I am just working out the last few kinks, namely:
- we've lost the concept of "script residency", i.e. we can't ask if a script is loaded currently, this is the reason teh integration tests didn't work with the asset loader, the "is loaded" condition was taken out of the loop
- fixing rhai bits and bobs
- something going on with the benchmarks

shanecelis:
Re: script residency. [Looks at the test harness lib.rs:305 file.] Oh, that's right. I imagine you're looking at this code:
```rust
        if let Some(event) = error_events.into_iter().next() {
            // eprintln!("XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXx");
            // if ! app.world().resource::<AssetServer>().load_state(&handle).is_loaded() {
            //     continue;
            // }
            return Err(event
                .error
                .display_with_world(WorldGuard::new_exclusive(app.world_mut())));
        }
```

I commented that out, but the script is being added a few lines above lib.rs:244.
```rust
    // The following code did not work, possibly because of the asynchronous
    // nature of AssetServer.
    //
    // ```
    // let handle = app.world_mut().resource_mut::<AssetServer>().load(&script_path);
    // app.world_mut().spawn(ScriptComponent::new([handle.clone()]));
    // ```
    let handle = {
        let mut script_dir = manifest_dir.clone();
        script_dir.push("assets");
        script_dir.push(script_id.path());
        // Read the contents and don't do anything async.
        let content = fs::read_to_string(&script_dir)
            .map_err(|io| format!("io error {io} for path {script_dir:?}"))?;
        let mut script = ScriptAsset::from(content);
        script.language = P::LANGUAGE;
        app.world_mut()
            .resource_mut::<Assets<ScriptAsset>>()
            .add(script)
    };
```
I was getting intermittent behavior if I recall correctly, so I figured if I could make script loading synchronous rather than async, it would be more reliable.

shanecelis:
Thanks! I appreciate the kind words. I'm happy to have you take the reigns. I felt like I had been pretty intrusive with my changes, so I tried not to upset the apple cart where I could, but I'm glad you're reconsidering some things like `Recipient`, which doesn't quite cut along nice lines like it used to. So if you can make the recipients cut along the new lines which will be performant, I think that's a good change.

The only things I'd want preserved are handles for scripts, in-order loading, and a notion of domains. (I realized with Nano-9 if I use a "nano-9" domain instead of a shared context then I wouldn't have to monopolize BMS, and the user would still be free to use BMS in any fashion they like, which feels like a win.)

I look forward to seeing your changes. Happy to test and review when it's ready.

makspll:
Alright a quick summary of the changes:
- Complete overhaul of integration tests, they are now driven by `scenario.txt` files which act akin to gherkin/specflow behaviour tests, scripts run against the nearest scenario file so you can have default scenarios run against most tests, and specialise for others
- `ScriptAttachment` concept, the user side of the context policy, each attachment is mapped to a `ContextKey`, right now the two attachments allowed are `ScriptEntity` and `StaticScript`. 
- Switch to using on_add and on_remove component hooks to emit attached detached events
- Removed the domain concept for now, as I didn't quite see how it would fit in here, having the domain as a separate component was a bit awkward (and it had to apply to all scripts on the script component which didn't quite feel right)
- Overhauled recipients and how the event handler searches for contexts, namely it no longer needs to care about unloaded scripts, as it will query direct from the context residents.
- I've experimented with instead of having a function variant letting people just use a selector trait implementation, I had to switch from Box to Arc though in the process when i needed Clone, so might revert that

PR #431, opened 2025-07-02
feat: Use handles with Bevy 0.16 support

shanecelis:
I felt bad about leaving you with a conflicted merge situation with my "use-handles" #427 with and @alexparlett's #422 "Update to Bevy 0.16", which I also want to see accepted, so I went ahead and merged @alexparlett's work on top of mine in this PR.

I think it'd be nice to have one last BMS release for Bevy 0.15 that includes #427 but that's your call.

I've intentionally left the code in a rough state. I haven't run `cargo format` or cleaned up the warnings because I wanted to make the issue of merging and rebasing less fraught. I wouldn't want to clean up #427 if a Bevy 0.15 release is not going to happen. 

So what is your release plan? Will another BMS for Bevy 0.15 be cut? Which PRs will go into it?


shanecelis:
So much has changed in the 0.15.3 branch for use-handles, I'm withdrawing this 0.16 PR in lieu of a forthcoming one.

PR #458, opened 2025-08-19
chore: update rust edition to 2024

makspll:
# Summary
About time!


makspll:
This is taking me on a journey, since some of the features in this edition require much newer nightly versions, and a version of rustc_plugin doesn't exist for those yet

PR #487, opened 2025-10-04
feat: Better HashMaps - hashmap iteration and read-access in map_get()

aggyomfg:
Hey! This PR improves how we work with HashMaps in the scripting layer. Two main things here:

### HashMap iteration support (commit b632625)
Added proper iteration over maps! Now you can do `for key, value in pairs(map)` in Lua and use `.iter()` in Rhai. This was trickier than it looks because Bevy's reflection path system doesn't play nice with maps - it rejects `ListIndex` access on map types. So I added a dedicated `ReflectMapRefIter` that bypasses the path system and uses `Map::get_at()` directly.

The iterator returns `(key, value)` tuples that get properly unpacked in both languages - Lua's `pairs()` just works, and Rhai gets a wrapped iterator with a `next()` method.

### Fixed `map_get()` to use read-only access (commit fc7d746)  
Changed `map_get()` from using `from_reflect()` to `to_dynamic()`. This might seem like a small thing but it's actually pretty important:

- **More reliable**: Works with ALL reflected types now, not just ones that have `ReflectFromReflect` registered. No more mysterious failures when accessing certain map value types.
- **Correct semantics**: Reading from a map doesn't need mutable access! Changed from `with_reflect_mut` to `with_reflect`. This is also groundwork for an upcoming PR that will fix Bevy's change detection by properly checking `ReflectReference` access modes.
- **Better perf**: Eliminates the registry lookup and type reconstruction that `from_reflect()` does. `to_dynamic()` is simpler.

## Testing

Added test scripts for both Lua and Rhai that verify:
- Iterating over map entries
- Getting correct key-value pairs
- Proper entry counts

Existing map operations should all still work as before.

### DIFF
--- a/assets/tests/iter/hashmap.lua
+++ b/assets/tests/iter/hashmap.lua
@@ -1,0 +1,15 @@
+local res_type = world.get_type_by_name("TestResourceWithVariousFields")
+local res = world.get_resource(res_type)
+
+local map = res.string_map
+
+local count = 0
+local found_keys = {}
+for key, value in pairs(map) do
+    count = count + 1
+    found_keys[key] = value
+end
+
+assert(count == 2, "Expected 2 entries, got " .. count)
+assert(found_keys["foo"] == "bar", "Expected foo=>bar")
+assert(found_keys["zoo"] == "zed", "Expected zoo=>zed")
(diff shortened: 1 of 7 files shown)


aggyomfg:
Added old map_get back as map_get_clone() - some times we still need to get fully Reflected values (for example can't constuct or dowcast asset handle from PartialReflect)

aggyomfg:
Adding map_get_clone() back pushed me to idea, that iters may also need fully reflected values, added this also.

PR #498, opened 2025-11-06
feat!: Bevy 0.17

makspll:
Tracking issues so far:
- [x] `WithWorldAccess` is not feasible anymore due to the inability to read component accesses into a system param at initialization
- [x] Handles don't have a weak wariant anymore, we wither go all AssetId or store strong handles, atm heading towards strong handles everywhere. This proves troublesome when receiveing modification events which don't have a handle, and you cannot get a strong handle with a path from an asset id (the path comes from the asset server load call, and not having it means script names don't show up in logs)
- [x] stack overflows with dynamic systems


makspll:
I've split out my holiday stash into a follow up: https://github.com/makspll/bevy_mod_scripting/pull/503, this PR should work as normal, but some of the bindings might not be optimal yet (like from_raw no longer exists without an optional wrapper, so codegen needs to improve to keep up with bevy) But the codegen improvements I am planning will drastically increase coverage to match existing trait implementations

shanecelis:
Yeah! Woohoo!

PR #529, opened 2026-03-28
feat!: extract `bevy_mod_scripting_world` decoupling world abstractions

makspll:
# Summary
The `bevy_mod_scripting_bindings` crate is bigger than it needs to, it would be ideal to have the ability to extend the world guard in separate crates, this PR is one step towards that.

I extracted the "cached" registry components into something less typed so that I could pull out `bevy_mod_scripting_world` which doesn't know about these, and instead alot of the functionality is pushed into `bevy_mod_scripting_bindings`, via the new `WorldExtensions` trait.

I also removed `ReflectAccessId` in favor of `WorldAccessRange` making the world crate a bit more opinionated on what kind of accesses exist


makspll:
Hmm not sure why the benches are alerting, when running locally this shows as an improvement, at least in the `Mut::from` benchmark:
image

makspll:
Claiming access for example looks to have sped up (probably because of the change from a dictionary to a list):
image

PR #533, opened 2026-04-15
fix: Improve pipeline robustness, allow same frame attach + detach events.

### LINKED ISSUE #468: :bug: Adding and remove script in the same function causes error
## Summary
I was trying to get a behaviour similar to before. I was using the fact that loading a script executed it to register scenarios to my app.

After my upgrade from BMS 0.13, I'm unsure how to reproduce the same behavior as before. I tried adding adding and removing the scenario on the same frame:

```rust
commands.queue(AddStaticScript::new(handle.clone()));
commands.queue(RemoveStaticScript::new(handle.clone()));
```

I got

```
2025-08-28T22:10:18.922027Z  WARN bevy_mod_scripting_core::commands: Attempted to delete static script id AssetId<bevy_mod_scripting_core::asset::ScriptAsset>{ index: 0, generation: 0}, but it was not found
```

## Reproducing
https://github.com/Ownezx/bms_min_repro/issues

makspll:
# Summary
This used to work in previous versions, but stopped when the new pipeline work got merged. Also simplifies things a bit.

This PR delays the "commitment" step of script machines further into the process. allowing both an attachment and detachment to happen in the same frame and be processed one after another.

This may be slightly jarring from a script perspective, as the entity you're attached to might be despawned by the time you execute, but that's unavoidable in this execution model.

Fixes #468

### DIFF
--- a/crates/bevy_mod_scripting_world/src/access_map.rs
+++ b/crates/bevy_mod_scripting_world/src/access_map.rs
@@ -316,7 +316,7 @@ impl AccessMapInner {
                     || self
                         .component_access
                         .iter_accessed()
-                        .any(|(_, a)| a.is_write() || (!a.is_write() && write))
+                        .any(|(_, a)| a.is_write() || write)
                 {
                     return false;
                 }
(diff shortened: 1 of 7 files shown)


crates/bevy_mod_scripting_world/src/access_map.rs
```diff
@@ -316,7 +316,7 @@ impl AccessMapInner {
                     || self
                         .component_access
                         .iter_accessed()
-                        .any(|(_, a)| a.is_write() || (!a.is_write() && write))
```
makspll, line 319: unrelated but small clearing up of logic

Issue #166, opened 2024-12-21
WASM Support


elliotsayes:
Trying to build for web gives:
`don't know how to build Lua for wasm32-unknown-unknown`
..originating from the `lua-src-rs` package. That package does support wasm, but only the `wasm32-unknown-emscripten` target which doesn't work with bevy

makspll:
As you see this is likely not possible until both mlua and bevy support a common wasm target

makspll:
> Looks like we'll get a WASM-compatible backend soon: [mastodon.world/@danielsreichenbach/116108388959479041](https://mastodon.world/@danielsreichenbach/116108388959479041)

Yup! Daniel has popped in to the bevy discord too, exciting!

julien-blanchon:
Hey @makspll any news about that. I need this feature in my projects and I'm currently using the danielsreichenbach fork (https://github.com/danielsreichenbach/bevy_mod_scripting/tree/feature/rilua-backend). Is this expected to come here anytime soon ?

makspll:
@julien-blanchon I am not currently planning on adding new language backends (i.e. rilua), I believe mlua should be able to support WASM as a target soon, which would mostly work, although I imagine we'd need to fix up some things in the codebase. In theory though if rilua supports WASM, an external crate could implement its scripting plugin without my help (would be curious to see what needs to change in the codebase if anything to compile fine on WASM)

ianm199:
Working on unblocking this. First step is getting the core to actually run on wasm — opened https://github.com/makspll/bevy_mod_scripting/pull/543 which fixes a std::time::Instant call in the pipeline that panics at runtime on wasm32-unknown-unknown (swaps to bevy_platform::time::Instant). The remaining blocker is the Lua backend's C dependency; I'm looking at a pure-Rust Lua VM for that next.

ianm199:
@makspll follow-up to my earlier comment. The external backend works now.

A pure-Rust Lua backend running in a Bevy wasm build: https://ianm199.github.io/bms-lua-rs/

It's Conway's Game of Life scripted in Lua, running on wasm32-unknown-unknown with no C and no emscripten. The script is bms's own game_of_life.lua, used as-is: it reads and writes the LifeState component through the reflection bridge (world.get_type_by_name, world.query, the component's cells). The page also lets you edit the Lua live and shows sliders the script declares for itself.

Some details:
[...]

PR #543, opened 2026-05-26
fix: Use bevy_platform::time::Instant in the script pipeline

ianm199:
`pipeline/machines.rs` uses `std::time::Instant`, which doesn't work on wasm:

- `std::time::Instant::now()` panics at runtime on `wasm32-unknown-unknown` ("time not implemented on this platform"); the machine tick loop calls it every frame.
- `bevy_platform::time::Instant` is already a dependency here — it's `std::time::Instant` on native and `web_time` on wasm. Same API, no behavior change off wasm, no new dep.

Towards #166.

### DIFF
--- a/crates/bevy_mod_scripting_bindings/src/function/script_function.rs
+++ b/crates/bevy_mod_scripting_bindings/src/function/script_function.rs
@@ -793,7 +793,7 @@ mod test {
             assert!(out.is_err());
 
             let gotten = out.unwrap_err();
-            let expected_function_name = "<bevy_mod_scripting_bindings::function::script_function::test::test_invalid_amount_of_args_errors_nicely::{{closure}} as bevy_mod_scripting_bindings::function::script_function::ScriptFunction<fn(usize, usize) -> usize>>::into_dynamic_script_function::{{closure}}";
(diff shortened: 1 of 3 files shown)


makspll (approved):

ianm199:
Pushed a commit to fix CI. It was red across the board, but not because of this change.

The benchmark and diff jobs fail while installing bencher, because bencher's current source uses `Duration::from_hours`, which is only const-stable on Rust 1.91 and up, while `rust-toolchain.toml` pins 1.89.0. The build and test matrix also trips on the pinned toolchain (clippy component missing) and then gets cancelled by fail-fast.

I bumped `rust-toolchain.toml` to 1.91.0 and declared the clippy and rustfmt components. Mind approving the workflow run so it can go again? Happy to split the toolchain bump into its own PR if you'd rather keep this one to just the `Instant` change.

ianm199:
One more commit. With the toolchain on 1.91, the only remaining failure was `test_invalid_amount_of_args_errors_nicely`: Rust 1.91 now renders the elided lifetime in `std::any::type_name`, so the expected `ScriptFunction<fn(..)>` is now `ScriptFunction<'_, fn(..)>`. Updated that hardcoded string. Benchmarks and bindings sync already went green on 1.91. Mind approving the run once more?

ianm199:
@makspll I confirmed this last one actually runs locally - sorry for all the back and forth. This should unblock other open PRs as well since the pipeline has been broke

makspll:
All good! Thanks for the contribution :D

Issue #545, opened 2026-06-11
Add a native Luau (.d.luau) backend for LAD files


jonasdedden:
# Add a native Luau (`.d.luau`) backend for LAD files

## Problem

`ladfile_builder` ships LAD post-processors for the Lua Language Server (`--- @class` `.lua`) and mdbook, but there's no native **Luau** one. `luau-lsp` can't consume the LuaLS dialect — it expects `declare class … end` / `declare name: T` syntax.

The `luau` runtime feature already exists, so you can *run* Luau scripts against Bevy, but there's currently no way to **type-check** them against the reflection registry.

## Proposal

A native Luau backend in `ladfile_builder`, behind a `luau_files` feature, mirroring how `lua_language_server_files` is wired:

- a pure `lad_to_luau(&LadFile, …) -> String` conversion, and
- a `LuauLadPlugin` implementing `LadFilePlugin`, added to `default_processors()`.

It would be a **feature-gated module** rather than a separate crate (like the LuaLS backend), since it's pure string generation over the `ladfile` types and needs **no new dependencies**.

Scope of a first cut:

- `declare class` for reflected types (with `focus_crates` to keep output readable), host global functions, and non-static instance handles like `world`;
- handle Luau grammar edges (reserved words, the unit type, static-vs-instance globals);
- unknown types resolve to `any`.

jonasdedden:
There now is `luau_lad_backend` (https://crates.io/crates/luau_lad_backend) externally, which could be used as a Luau backend.

PR #546, opened 2026-06-11
feat(ladfile_builder): First look into a native Luau (.d.luau) LAD backend

jonasdedden:
This is a first proposal for generating Luau type definition files (corresponding issue: https://github.com/makspll/bevy_mod_scripting/issues/545). Please let me know whether you like the general approach or not.

## Tests

`src/lib.rs` carries module tests that parse a real registry dump (`src/test_assets/bindings.lad.json`) and assert both modes:

- `converts_general_surface` — classes/fields/`world`/host globals present, and **no** `Reg<T>` leakage in the default output.
- `converts_branded_surface` — `Reg<T>`, the generic `get_component`, and the `…Type` handle branding all present when opted in.
- `sanitizes_and_escapes` — reserved-word handling.

Plus a well-formedness sweep asserting no Luau keyword is emitted as a bare field/method/param name. `cargo test` in this folder runs them against `ladfile`.

## Luau grammar edges handled

- reserved words can't be bare identifiers — quoted as `["end"]` for fields, suffixed (`end_`) for methods/params/type names;
- the unit type is `()` only in a return arrow — rendered as `nil` elsewhere;
- BMS exposes every reflected type as a *static* accessor global — `is_static` separates those from real instance handles like `world`;
- unknown/foreign types resolve to `any` (Luau treats it permissively), kept focused via `focus_crates`.

### DIFF
--- a/crates/ladfile_builder/CHANGELOG.md
+++ b/crates/ladfile_builder/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Native **Luau** (`.d.luau`) LAD backend (`luau` module, behind the new
+  `luau_files` feature) so Luau scripts can be type-checked with `luau-lsp`.
+  Exposes `lad_to_luau` (a pure `&LadFile -> String` conversion) and
+  `LuauLadPlugin` (a `LadFilePlugin` processor wired into `default_processors()`).
+  Brings no new dependencies. An opt-in `HandleBranding` configuration emits a
+  `Reg<T>` phantom brand for cast-free, statically-typed component access.
+
 ## [0.19.0](https://github.com/makspll/bevy_mod_scripting/compare/ladfile_builder-v0.18.0...ladfile_builder-v0.19.0) - 2026-01-25
 
 ### Added
--- a/crates/ladfile_builder/Cargo.toml
+++ b/crates/ladfile_builder/Cargo.toml
@@ -15,6 +15,9 @@ readme.workspace = true
 [features]
 default = []
 lua_language_server_files = ["lua_language_server_lad_backend"]
+# Native Luau (.d.luau) definition-file backend. Pure string generation over the
+# LAD types, so it needs no extra dependency.
+luau_files = []
 
 [dependencies]
 bevy_app = { workspace = true, default-features = false, features = [] }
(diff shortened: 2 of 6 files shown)


makspll (commented):
Hi, I had a look and I can see at least a couple problems:
- the whole implementation lives in `ladfile_builder`, that crate serves to purely create ladfiles not provide language specific implementations, the whole implementation should live in its own crate
- Some conventions here are somewhat opinionated and require non-default bindings (branding for example), generally in LAD backends this is not the case and type code only describes what already exists in BMS
- Escaped identifiers (with underscore suffix) don't have support in the LUA backend in BMS, so even though you type them this way they won't have anything backing them.
- Hardcoding types like `World` and `Entity`, these do also live in ladfiles so backends should be able to deal with them fully dynamically with some minor exceptions (for example ladfiles store 'is_core' information which is also user customizable, which drives type priority in generated documentation for example)


Overall especially since I am not familiar with luau too much, I'd suggest the following:
- Since new LAD backends don't require to live in the BMS crate, this can exist in its own separate crate fully
- Once it has gained traction and many BMS consumers are using it, I would consider then merging the crate into the BMS project after all the kinks are worked out

Ownezx:
I was following this to see where it would go. I've been using Luau in my project and I am interested in creating LAD files for it.

If you do create a new crate with the LAD backend please let me know, I'd be happy to give it a go and give feedback. If I have the energy I might also contribute.

jonasdedden:
Hey @makspll and @Ownezx,

Thanks for your feedback! I'm happy to extract this into a separate crate that I can bring upstream on my own for now. :) The current status is just a reflection of the bare minimum vibecoded status that I'm using right now locally to enable it in general, happy to clean it up based on the remarks you gave.

I really want to prevent name collisions and unwarranted connections to your project: would you be fine with the name `bevy_mod_scripting_luau`? If you would be happy with a different crate name, please let me know!

jonasdedden:
Hey @Ownezx,

I worked a bit further on this idea and created a crate here that you could check out: https://github.com/jonasdedden/luau_lad_backend

If the name is okay, and people find it interesting, I would publish it on crates.io.

EDIT: I also have some tiny tower defense game that shows how this crate could be used in a larger context: https://github.com/jonasdedden/bevy-typed-luau-tower-defense

Ownezx:
Maybe we can use it as a repo dependency for now? I have still not had time to check as I've been trying to get some networking on my side and will still be for a little bit...

jonasdedden:
I now published `luau_lad_backend` (https://crates.io/crates/luau_lad_backend) externally, so I will close this PR.

